PR #87, opened 2023-01-20
XLS to rippled amendment guideline proposal

Silkjaer:
Suggesting more explicit guidelines for XLS drafts that leads to rippled amendments.


intelliot (approved):

Silkjaer:
As discussed elsewhere, we may want to move drafts to be code PR's, pending as PR proposals until the standard is no longer a draft. This will allow edit proposals and line-by-line comments etc.

PR #104, opened 2023-04-19
Adding XLS-39d-Clawback specification

shawnxie999:
This is the XLS-39d spec proprosal for what originally is in [discussion](https://github.com/XRPLF/XRPL-Standards/discussions/94).

The major spec change from the original proposal is that `lsfNoClawback` flag is removed from the `AccountRoot`. Now, Clawback is purely an extension of freeze, such that a trustline can only be clawed back if and only if the trustline has been frozen. 

This spec if open to discussion about whether we would still want a seperate `lsfNoClawback` flag to allow the user to disable clawback entirely.

### DIFF
--- a/XLS-39d-Clawback/README.md
+++ b/XLS-39d-Clawback/README.md
@@ -1,0 +1,149 @@
+<pre>
+  Title:        <b>Clawback Support</b>
+  Revision:     <b>2</b> (2023-4-19)
+
+<hr>  Author:       <a href="mailto:nikb@bougalis.net">Nikolaos D. Bougalis</a>
+                <a href="mailto:shawnxie@ripple.com">Shawn Xie</a>
+</pre>
+
+# 1. Clawback Support
+
+## 1.1. Abstract
+
+Although the XRP Ledger offers rich support for [tokens](https://xrpl.org/tokens.html) (a.k.a. IOUs or issued assets), including offering issuers the ability to [freeze](https://xrpl.org/freezes.html) issuances, in order to meet regulatory requirements, some issuers need to be able to go further, by having the ability to "[clawback](https://en.wikipedia.org/wiki/Clawback)" their issued assets. 
+
+----------------------- -------------------------------------------------------
+:bangbang: This proposal deals only with issued assets. **The proposed clawback
+           support _cannot_ be used to claw back XRP.**                                        
+----------------------- -------------------------------------------------------
+
+### 1.1.1. Advantages and Disadvantages
+
+**Advantages**
+
+- Issuers that are restricted from issuing on ledger because of regulatory requirements requiring the ability to clawback funds will be able to issue tokens.
+- By being able to "clawback" issuers can ensure that the "on-chain" view is representative of the balance.
+- Compared to other on-ledger features (e.g. freeze), clawback is minimal and trivial to implement.
+- Clawback is implemented as an extension of the existing freeze feature, meaning that clawback can be toggled on/off from an `TrustLine` level. It also makes sense to bundle these two features to serve specific regulatory needs.
+- By relying on the freeze flag, Clawback ensures that issuers who have enabled `lsfNoFreeze` will continue to be unable to access holder's funds, even after the introduction of the clawback feature. This reinforces trust among token holders, as they can be confident that their funds will remain secure and untouched.
+
+
+**Disadvantages**
+
+- Introduces an additional transaction, along with the complexity that comes with that.
+- Issuers now get additional power, which may be concerning to token holders.
+- Requires additional documentation, including highlighting the fact that clawback cannot be applied to XRP.
+
+## 1.2. Creating and Transferring Tokens on XRPL
+
+### 1.2.1. On-Ledger Data Structures
+
+This proposal introduces no new on ledger structures.
+
+## Transactions
+
+This proposal introduces one new transaction: `Clawback`:
+
+### The **`Clawback`** transaction
+
+The **`Clawback`** transaction modifies a trustline object, by adjusting the balance accordingly and, if instructed to, by changing relevant flags. If possible (i.e. if the `Clawback` transaction would leave the trustline is in the "default" state), the transaction will also remove the trustline.
+
+**An counterparty can be clawed back if and only if the trustline has been frozen.** There are two ways to enable freeze on a trustline, either through a `TrustSet` or `Clawback` where the `Flags` field is set accordingly. If a clawback transaction is attempted on a trustline that has not been frozen, the transaction will not be allowed and will return with an error code `tecNO_PERMISSION`.
+
+The transaction supports all the existing "common" fields for a transaction.
+
+#### Transaction-specific Fields
+
+| Field Name          | Required?        |  JSON Type    | Internal Type     |
+|---------------------|:----------------:|:-------------:|:-----------------:|
+| `TransactionType`   |:heavy_check_mark:|`string`       |   `UINT16`        |
+
+Indicates the new transaction type **`Clawback`**. The integer value is `30`. The recommended name is `ttCLAWBACK`.
+
+| Field Name    | Required?        |  JSON Type  | Internal Type     |
+|---------------|:----------------:|:-----------:|:-----------------:|
+| `Account`     |:heavy_check_mark:|`string`     | `ACCOUNT ID`      |
+
+Indicates the account which is executing this transaction. The account **MUST** be the issuer of the asset being clawed back. Note that in the XRP Ledger, trustlines are bidirectional and, under some configurations, both sides can be seen as the "issuer" of an asset. In this specification, the term issuer is used to mean the side of the trustline that has an outstanding balance (i.e. 'owes' the issued asset) that it wishes to claw back.
+
+---
+
+| Field Name | Required?        | JSON Type | Internal Type |
+|------------|:----------------:|:---------:|:-------------:|
+| `Amount`   |:heavy_check_mark:| `object`  |   `AMOUNT`    |
+
+Indicates the amount being clawed back, as well as the counterparty from which the amount is being clawed back from. It is not an error if the amount exceeds the holder's balance; in that case, the maximum available balance is clawed back. It is not an error if the amount is zero; in this case, the transaction does not claw back any funds but may adjust flags, as indicated by the `Flags` field.
+
+It is an error if the counterparty listed in `Amount` is the same as the `Account` issuing this transaction; the transaction should fail execution with `temBAD_AMOUNT`.
+
+:bangbang: The sub-field `issuer` within `Amount` represents the token holder's address instead of the issuer's.
+
+---
+
+| Field Name     | Required?        |  JSON Type  | Internal Type     |
+|----------------|:----------------:|:-----------:|:-----------------:|
+| `Flags`        |:heavy_check_mark:|`number`     | `UINT32`          |
+
+Specifies the flags for this transaction. The universal transaction flags that are applicable to all transactions (e.g., `tfFullyCanonicalSig`) are valid. This proposal introduces the following new, transaction-specific flags:
+
+>| Flag Name     |         Flag Value         |                                                         Description                                                         |
+>|---------------|:---------------------------------------------------------------------------------------------------------------------------:|:----------------------------------------------------------------------------------------------------------------------------------------------------------------------:|
+>| `tfSetFreeze`    |        `0x00000001`        | If set, either the `lsfHighFreeze` or `lsfLowFreeze` flag (as appropriate) will be set in the trust line. It is not an error to request to set a flag that has already been set. Enabling the `tfSetFreeze` flag simplifies the process of clawing back funds from a trustline. It allows the clawback transaction to freeze the trustline and reclaim the funds in a single transaction, without needing to set the `tfSetFreeze` flag through a `TrustSet` transaction beforehand. However, if the `tfSetFreeze` flag has not been set or the appropriate freeze flag hasn't been enabled on the trustline before attempting to clawback, then the transaction will fail and return the error code `tecNO_PERMISSION`.|
+>| `tfClearFreeze`      |        `0x00000002`        | If set, either the `lsHighFreeze` or `lsfLowFreeze` flag (as appropriate) will be removed from the trust line. It is not an error to request to clear a flag that isn't set. Enabling the `tfClearFreeze` flag simplifies the process of unfreezing a trustline. With this flag enabled, the clawback transaction can reclaim the frozen funds and unfreeze the trustline at the same time, without requiring an additional `TrustSet` transaction to clear the freeze flag.|
+
+The `tfSetFreeze` and `tfClearFreeze` flags are options that can be set in a `Clawback` transaction. They make the process of clawing back funds more convenient by combining multiple steps into one.
+
+Usually, when you want to clawback funds from a trustline, you have to freeze the trustline first, then take back the funds, and finally, unfreeze the trustline if you choose to. With these two flags, you can clawback the funds and keep the trustline unfrozen at the same time by setting both of them simultaneously.
+
+---
+
+## Example **`Clawback`** transaction
+
+```
+{
+  "TransactionType": "Clawback",
+  "Account": "rp6abvbTbjoce8ZDJkT6snvxTZSYMBCC9S",
+  "Amount": "LimitAmount": {
+      "currency": "FOO",
+      "issuer": "rsA2LpzuawewSBQXkiju3YQTMzW13pAAdW",
+      "value": "314.159"
+    },
+  "Flags": 1,
+  "Fee": 10,
+  "Memos": [
+        {
+            "Memo": {
+                "MemoType": "526561736f6e",
+                "MemoData": "5468697320636c61776261636b2077617320617574686f72697a6564206279204a6f652e"
+            }
+        }
+    ],
+}
+```
+
+#### Execution
+
+In execution, this transaction would freeze the trustline by setting the `Flag` to 1, and claw back at most **314.159 FOO** issued by `rp6abvbTbjoce8ZDJkT6snvxTZSYMBCC9S` and held by `rsA2LpzuawewSBQXkiju3YQTMzW13pAAdW`. If `rsA2LpzuawewSBQXkiju3YQTMzW13pAAdW` did not have a trustline set up or that trustline was set to `0` then the error `tecNO_LINE` would be returned and a fee would be consumed.
+
+### 1.3. Account Root modifications
+
+This proposal does not introduce any new flags for the clawback functionality. But, if the issuer has already set the `lsfNoFreeze` flag, then it means that they will not be able to clawback either.
+
+### 1.4 Amendment
+
+This transaction will require an amendment. The proposed name is `XLS-39-Clawback`.
+
+### 1.5 Compatibility with Compact Fungible Token (XLS-33d)
+The Compact Fungible Token(CFT) proposes another token type. The clawback functionality can be extended to CFT because it supports the freeze feature, just like trustline. With CFT, the issuer can choose to freeze individual token holder's balance. If clawback is updated in order to accomodate for this new token, the `Clawback` transaction introduces a new field named `CFTokenAmount`:
+
+---
+
+| Field Name | Required?        | JSON Type | Internal Type |
+|------------|:----------------:|:---------:|:-------------:|
+| `CFTokenAmount`   || `object`  |   `CFTOKEN AMOUNT`  |
+
+The `CFTokenAmount` encompasses `CFTokenIssuanceID` and `Value` to indicate the CFT and the amount to be clawed back.
+
+With the addition of this field, the `Amount` field is now *optional*. But, either one of `CFTokenAmount` or `Amount` must be specified in the `Clawback` transaction, otherwise `temMALFORMED` is returned from the transaction.
+
+---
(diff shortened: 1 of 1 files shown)


Silkjaer:
Title in the filename should be lowercase and the draft should use the template format: https://github.com/XRPLF/XRPL-Standards/blob/master/xls-template.md

Also, for good measure, the author of the proposal should move this from discussion to draft, but if he is OK with it in the current form, let's progress 👍

CC: @nbougalis

nbougalis:
I'm fine with @shawnxie999 taking over this proposal.

nbougalis:
>We spoke with stakeholders and they said they can't think of a scenario where funds are clawed back without freezing a trustline, unless we can come up with any. Normally, freezing is always a precursor of clawback in real world use case.

Perhaps _operationally_ that's how it's done, but the spec shouldn't present this as a de-facto _technical requirement_.

shawnxie999:
@nbougalis 
> Perhaps _operationally_ that's how it's done, but the spec shouldn't present this as a de-facto _technical requirement_.

If this isn't part of the technical requirement, it will give more flexibility for the issuer. But, I think it would be more important to care about what the token holder think when it comes to the clawback ability. When this amendment goes live, token holders will be _really_ concerned knowing that the issuer can randomly clawback all of their funds.

XLS-39d-clawback/README.md
```diff
@@ -0,0 +1,168 @@
...
+
+---
+
+#### 3.2.2. Example **`Clawback`** transaction
+
+```
+{
+  "TransactionType": "Clawback",
+  "Account": "rp6abvbTbjoce8ZDJkT6snvxTZSYMBCC9S",
+  "Amount": "LimitAmount": {
```
sappenin, line 112: Is this a typo?

XLS-39d-clawback/README.md
```diff
@@ -0,0 +1,168 @@
...
+>| `tfClearFreeze`      |        `0x00000002`        | If set, either the `lsHighFreeze` or `lsfLowFreeze` flag (as appropriate) will be removed from the trust line. It is not an error to request to clear a flag that isn't set. Enabling the `tfClearFreeze` flag simplifies the process of unfreezing a trustline. With this flag enabled, the clawback transaction can reclaim the frozen funds and unfreeze the trustline at the same time, without requiring an additional `TrustSet` transaction to clear the freeze flag.|
```
sappenin, line 98: Is it expected that an issuer would clawback funds, but then unfreeze the trustline to allow the offending party to continue transacting?

shawnxie999: Yes, if the issuer thinks the clawback action will resolve their concern with the offending party, then they can set the `tfClearFreeze` flag in the Clawback transaction to unfreeze the trustline at the same time. But if the issuer wants to keep the trustline frozen, they can just leave this flag out.

XLS-39d-clawback/README.md
shawnxie999: Yes, will fix

shawnxie999:
I've updated the spec to add an additional `lsfNoClawback` flag to give more flexibility, though it's still serving as an extension of freeze. Although it would be desirable to have clawback disabled by default, it would be best to follow what has already been put in place with the default freeze settting(enabled by default). I've added an `Rationale` section to explain this.

shawnxie999:
Updated the spec again to reflect the change of the global flag: from `lsfNoClawback` to `lsfAllowClawback`, so that clawback is opt-out by default. We don't we to keep proposing new features that are enabled by default to increase concerns of the token holders. The existing freeze feature is an exception. From now on, any feature of this nature should be opt-out by default.

scottschurr (commented):
By the time I arrived at the bottom of the spec it seemed that requiring clawback to only occur on a frozen trust line was silly.  That requirement is all too easy for an issuer to work around.

The restrictions regarding the `lsfAllowClawback` flag on the issuer seem reasonable and sufficient.  And I agree that the proposed relationship between the `lsfAllowClawback` and `lsfNoFreeze` flags seems good.

I'll be curious to hear counter arguments.

XLS-39d-clawback/README.md
```diff
@@ -0,0 +1,182 @@
...
+
+This proposal introduces 1 additional flag for the `Flags` field of `AccountRoot`:
+
+| Flag Name       |  Flag Value  |
+|:---------------:|:------------:|
+| `lsfAllowClawback` | `0x02000000` | 
+
+Clawback is disabled by default. The account must set this flag through an `AccountSet` transaction, which is successful only if the account has an empty owner directory, meaning they have no trustlines, offers, escrows, payment channels, or checks. Otherwise, the `AccountSet` returns `tecHAS_OBLIGATIONS`. After this flag has been successfully set, it cannot reverted, and the account permanently gains the ability to clawback on frozen trustlines.
+
+If the account attempts to set `lsfAllowClawback` while `lsfNoFreeze` is set, the transaction will return `tecNO_PERMISSION` because clawback cannot be enabled on an account that has already disclaimed the ability to freeze trustlines.
```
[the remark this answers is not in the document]
shawnxie999: Thanks! updated.

XLS-39d-clawback/README.md
```diff
@@ -0,0 +1,182 @@
...
+
+
+### 3.3. Transactions
+This proposal introduces one new transaction: `Clawback`
+
+#### 3.3.1. `Clawback` transaction
+The **`Clawback`** transaction modifies a trustline object, by adjusting the balance accordingly and, if instructed to, by changing relevant flags. If possible (i.e. if the `Clawback` transaction would leave the trustline is in the "default" state), the transaction will also remove the trustline.
+
+**An counterparty can be clawed back if and only if the trustline has been frozen and `lsfAllowClawback` is set.** There are two ways to enable freeze on a trustline, either through a `TrustSet` or `Clawback` where the `Flags` field is set accordingly. If a clawback transaction is attempted on a trustline that has not been frozen, the transaction will not be allowed and will return with an error code `tecNO_PERMISSION`.
```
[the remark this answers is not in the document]
shawnxie999: Thanks for the callout. I implicitly assumed a global freeze would freeze individual trustlines, perhaps I wasn't being clear. Upated.

XLS-39d-clawback/README.md
```diff
@@ -0,0 +1,182 @@
...
+>| `tfClearFreeze`      |        `0x00000002`        | If set, either the `lsHighFreeze` or `lsfLowFreeze` flag (as appropriate) will be removed from the trust line. It is not an error to request to clear a flag that isn't set. Enabling the `tfClearFreeze` flag simplifies the process of unfreezing a trustline. With this flag enabled, the clawback transaction can reclaim the frozen funds and unfreeze the trustline at the same time, without requiring an additional `TrustSet` transaction to clear the freeze flag.|
+
+The `tfSetFreeze` and `tfClearFreeze` flags are options that can be set in a `Clawback` transaction. They make the process of clawing back funds more convenient by combining multiple steps into one.
+
+Usually, when you want to clawback funds from a trustline, you have to freeze the trustline first, then take back the funds, and finally, unfreeze the trustline if you choose to. With these two flags, you can clawback the funds and keep the trustline unfrozen at the same time by setting both of them simultaneously.
```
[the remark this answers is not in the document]
shawnxie999: Clawback and freeze are bundled is because both are meant for regulatory compliance, where freezing a trustline is usually the first step when the issuer notices suspicious activity and then clawback follows. 

Introducing a standalone clawback feature, independent of freeze, can cause concern among holders, particularly they don't know in what ways the issuer will use it. By bundling clawback with freeze, we can provide reassurance for holders that this clawback feature is to be used with the existing freeze feature, which is something they already acknowledge. This also brings a better communication strategy when introducing clawback - we would not be scaring people away by introducing an _standalone_ feature. 

>To me it looks like this completely obviates the need for Clawback to be done on a frozen trust line. It's trivial for the issuer to work around that requirement.

The fact that issuer can easily workaround doesn't conflict with the idea of trying to make users feel safer. I don't see of downside of bundling the clawback with freeze when we can gain a lot benefits, for the reasons stated above. This design choice is made with the best interests of token holders in mind. Yes, issuers may find it silly since they can easily workaround it, but this is okay and not within the consideration of token holders.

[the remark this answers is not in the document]
scottschurr: Your argument for tying clawback to account freezing is marketing based, not engineering based.  I'm an engineer, not a marketing person.  From an engineering perspective, I think tying clawback to account freeze is pointless.  I don't hear you disagreeing with that.

I'm in no sense a marketing person.  So there may be legit marketing reasons to tie those features together.  But I think there is marketing risk as well.  The risk would come in three phases:

1. Some folks promote that clawback is hard for an issuer to abuse because it is tied to trust line freezing.
2. Some (_any_) issuer abuses clawback (easily) because trust line freezing in actuality presents no barrier.
3. Users of the XRP Ledger learn to mistrust those who said that clawback is hard to abuse.

Promoting things that look like safety precautions, but don't actually increase safety, can backfire.  That's worth including in the marketing calculus.

The _real_ safety measures with clawback are two fold:

1. The technical safety measure is that the `lsfAllowClawback` flag must be set on an issuing account before any trust lines are created.  This allows the counterparty of each trust line to decide for themselves whether they trust the issuer not to abuse clawback.
2. The other safety measure is social.  Any issuer that casually performs clawback without strong justification will lose the trust and goodwill of many users of their currency.  Such risks should not be taken lightly.

My belief is that the XRP Ledger should stick to _actual_ safety measures.  Any pretend safety measures risk backfiring.

In summary,

1. I'd prefer that the XRP Ledger not promote fake safety measures.  It's dishonest and can backfire.
2. If we're not going to promote fake safety measures, then let's not include them in our product design.

XLS-39d-clawback/README.md
```diff
@@ -0,0 +1,182 @@
...
+
+`Clawback` is incorporated into freeze because, from the perspective of the token holder, the impact of `Clawback` is comparable to that of freeze. In particular, token holders who have already frozen trustlines would not be concerned by the introduction of the `Clawback` feature, as their funds are already inaccessible. Therefore, from the perspective of the holder, the presence or absence(after claw back) of funds in their frozen trustline is unsignficant. By tying `Clawback` to freeze, issuers can minimize the potential impact on token holders and avoid creating unnecessary concern. This approach provides issuers with greater flexibility and control over their assets, while also preserving the trust and confidence of token holders in the ecosystem.
```
ximinez, line 45: Nit: "unsignificant" -> "insignificant"

XLS-39d-clawback/README.md
```diff
@@ -0,0 +1,182 @@
...
+
+
+
+### 3.2. Account Root modifications
+
+This proposal introduces 1 additional flag for the `Flags` field of `AccountRoot`:
+
+| Flag Name       |  Flag Value  |
+|:---------------:|:------------:|
+| `lsfAllowClawback` | `0x02000000` |
```
ximinez, line 63: This value is reserved for Hooks. https://github.com/XRPLF/rippled/blob/develop/src/ripple/protocol/LedgerFormats.h#L236

XLS-39d-clawback/README.md
```diff
@@ -0,0 +1,182 @@
...
+
+---
+
+| Field Name | Required?        | JSON Type | Internal Type |
+|------------|:----------------:|:---------:|:-------------:|
+| `Amount`   |:heavy_check_mark:| `object`  |   `AMOUNT`    |
+
+Indicates the amount being clawed back, as well as the counterparty from which the amount is being clawed back from. It is not an error if the amount exceeds the holder's balance; in that case, the maximum available balance is clawed back. It is not an error if the amount is zero; in this case, the transaction does not claw back any funds but may adjust flags, as indicated by the `Flags` field.
+
+It is an error if the counterparty listed in `Amount` is the same as the `Account` issuing this transaction; the transaction should fail execution with `temBAD_AMOUNT`.
```
ximinez, line 102: The example in 3.3.2 also indicates that it is an error if no trustline exists, or the trustline is set to 0 (does this mean the balance is 0?). I think it would make more sense to have that specified here along with the other possible errors, instead of as part of an example.

ximinez (changes requested):

XLS-39d-clawback/README.md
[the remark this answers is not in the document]
shawnxie999: > The _real_ safety measures with clawback are two fold:
> 
> 1. The technical safety measure is that the `lsfAllowClawback` flag must be set on an issuing account before any trust lines are created.  This allows the counterparty of each trust line to decide for themselves whether they trust the issuer not to abuse clawback.
> 2. The other safety measure is social.  Any issuer that casually performs clawback without strong justification will lose the trust and goodwill of many users of their currency.  Such risks should not be taken lightly.

For context, initially, clawback was proposed to be opt-in by default, introducing a certain level of risk, which was why we thought bundling clawback with freeze would be a good idea (so that the issuer can't clawback randomly). Since having clawback enabled by default was really "dangerous", we tied the two features to ensure freeze is a prerequisite to clawback.

Now, with the latest iteration of the spec, clawback is opt-out by default, and has limitation to owner directory, which is a huge safety measure itself. So I do agree that tying clawback with freeze doesn't provide many benefits as before. 

>Your argument for tying clawback to account freezing is marketing based, not engineering based

I don't think we are going to promote the "tying freeze with clawback" as a safety precaution, or do any marketing of that kind. T made sense to do so since both are meant to serve regulatory compliance, hence we engineer the feature around real-world use cases. 

But, I do think it is okay to untie clawback with freeze, and don't have a strong argument to counter that. I'll wait for others to see their opinion.

XLS-39d-clawback/README.md
```diff
@@ -0,0 +1,178 @@
...
+If the account attempts to set `lsfAllowClawback` while `lsfNoFreeze` is set, the transaction will return `tecNO_PERMISSION` because clawback cannot be enabled on an account that has already disclaimed the ability to freeze trustlines. Reversely, if an account attempts to set `lsfNoFreeze` while `lsfAllowClawback` is set, the transaction will also return `tecNO_PERMISSION`.
+
+
+### 3.3. Transactions
+This proposal introduces one new transaction: `Clawback`
+
+#### 3.3.1. `Clawback` transaction
+The **`Clawback`** transaction modifies a trustline object, by adjusting the balance accordingly and, if instructed to, by changing relevant flags. If possible (i.e. if the `Clawback` transaction would leave the trustline is in the "default" state), the transaction will also remove the trustline.
+
+**An counterparty can be clawed back if and only if `lsfAllowClawback` is set.** If this transaction is attempted while `lsfAllowClawback` is unset, it will return with an error code `tecNO_PERMISSION`.
```
sappenin, line 72:
```suggestion
**Issued tokens can be clawed back by an issuer if, and only if, `lsfAllowClawback` is set.** If this transaction is attempted while `lsfAllowClawback` is unset, it will return with an error code `tecNO_PERMISSION`.
```

XLS-39d-clawback/README.md
```diff
@@ -0,0 +1,178 @@
...
+  Revision:     <b>5</b> (2023-5-08)
+<hr>  Author:       <a href="mailto:nikb@bougalis.net">Nikolaos D. Bougalis</a>
+                <a href="mailto:shawnxie@ripple.com">Shawn Xie</a>
+<hr>  Requires:     XLS 39
+<hr>  core_protocol_changes_required:     true
+</pre>
+
+## 1. Abstract
+
+Although the XRP Ledger offers rich support for [tokens](https://xrpl.org/tokens.html) (a.k.a. IOUs or issued assets), including offering issuers the ability to [freeze](https://xrpl.org/freezes.html) issuances, in order to meet regulatory requirements, some issuers need to be able to go further, by having the ability to "[clawback](https://en.wikipedia.org/wiki/Clawback)" their issued assets.  **An counterparty can be clawed back if and only if `lsfAllowClawback` is set.**
```
sappenin, line 13:
```suggestion
Although the XRP Ledger offers rich support for [tokens](https://xrpl.org/tokens.html) (a.k.a. IOUs or issued assets), including offering issuers the ability to [freeze](https://xrpl.org/freezes.html) issuances, in order to meet regulatory requirements, some issuers need to be able to go further, by having the ability to "[clawback](https://en.wikipedia.org/wiki/Clawback)" their issued assets.  **Issued tokens can be clawed back by an issuer if and only if `lsfAllowClawback` is set.**
```

XLS-39d-clawback/README.md
```diff
@@ -0,0 +1,178 @@
...
+
+### 3.4. Amendment
+
+This transaction will require an amendment. The proposed name is `XLS-39-Clawback`.
+
+---
+
+## 4. Rationale
+
+Clawback is disabled by default and requires the issuer to have an empty owner directory(no existing trustlines, offers, etc) before they are allowed to enable the feature. This is meant to be a conservative design, where the issuer needs to go through the major effort in order to enable this feature. This process is a huge safety measure to ensure the holders are aware of this feature if the the issuer wants to enable it.
```
sappenin, line 156:
```suggestion
Clawback is disabled by default and requires the issuer to have an empty owner directory (i.e., no existing trustlines, offers, etc) before the issuer is allowed to enable this feature. This design prefers a cautious approach where the issuer needs to take overt action to enable clawback capabilities. This helps ensure that token holders are aware of the possibility of clawback before they choose to hold any particular token.
```

XLS-39d-clawback/README.md
```diff
@@ -113,12 +111,11 @@ Specifies the flags for this transaction. The universal transaction flags that a
...
+>| `tfSetFreeze`    |        `0x00000001`        | If set, either the `lsfHighFreeze` or `lsfLowFreeze` flag (as appropriate) will be set in the trust line. It is not an error to request to set a flag that has already been set.|
+>| `tfClearFreeze`      |        `0x00000002`        | If set, either the `lsHighFreeze` or `lsfLowFreeze` flag (as appropriate) will be removed from the trust line. It is not an error to request to clear a flag that isn't set. |
 
-The `tfSetFreeze` and `tfClearFreeze` flags are options that can be set in a `Clawback` transaction. They make the process of clawing back funds more convenient by combining multiple steps into one.
+The `tfSetFreeze` and `tfClearFreeze` flags are options that can be set in a `Clawback` transaction. They make the process more inconvinient in case the issuer wants to clawback and freeze at the same time.
```
ximinez, line 117: Typo: `inconvinient` -> `convenient`

ximinez (changes requested):

XLS-39d-clawback/README.md
shawnxie999: will change. thanks!

XLS-39d-clawback/README.md
shawnxie999: will change. thanks!

XLS-39d-clawback/README.md
shawnxie999: changed

XLS-39d-clawback/README.md
shawnxie999: fixed

XLS-39d-clawback/README.md
ximinez: Yyyyyyyyyup......

XLS-39d-clawback/README.md
```diff
@@ -0,0 +1,178 @@
+<pre>
+  Title:        <b>Clawback Support</b>
+  Description:  Extending clawback functionality into freeze
+  Revision:     <b>5</b> (2023-5-08)
+<hr>  Author:       <a href="mailto:nikb@bougalis.net">Nikolaos D. Bougalis</a>
+                <a href="mailto:shawnxie@ripple.com">Shawn Xie</a>
+<hr>  Requires:     XLS 39
+<hr>  core_protocol_changes_required:     true
+</pre>
```
Silkjaer, line 9: Should be updated to reflect the xls template. Revision is not used anymore (small standard revisions will be backtracable through git history, major revisions (post XLS) will require an XLS standard number bump)

shawnxie999:
The latest revision removed the flags for clawback transaction.

XLS-39d-clawback/README.md
```diff
@@ -0,0 +1,169 @@
...
+- The account that signs and submits `Clawback` transaction must be the token issuer 
+- Token issuer cannot clawback from themselves
+- `Clawback` adheres to account flags `lsfAllowClawback`, `lsfGlobalFreeze` and `lsfNoFreeze`
+- The issuer is only able to claw back the specific amount of funds that specified in the transaction, but can't exceed the maximum amount of funds the holder has
+- Test that the `Clawback` feature does not interfere with any other features of the token, such as Offers
+
+## 7. Compatibility with Automated Market Maker (XLS-30)
+The Automated Market Maker (AMM) gives an account the ability to deposit issued tokens into AMM instance pool, in the form of `LPToken`. As of the current `Clawback` spec, it only allows an issuer to claw back the funds that are _spendable_. This would mean that the funds deposited into the AMM pool cannot be clawed back.
+
+If clawing back from AMM instance pool is something needed, this change will need a separate specification to address its feasibility, after AMM goes live on mainnet.
```
intelliot, line 169:
```suggestion
If clawing back from an AMM instance pool is required, such change will need a separate specification.
```

intelliot (approved):

intelliot (approved):

scottschurr (approved):
Spec looks great!  Simple and direct.

I spotted a couple of editing kinds of things that might want to be cleaned up.  But nothing that should hold up release.

Silkjaer (approved):

lathanbritz:
This really is the first issue that has _highlighted the interdependency_ that is some times present in tickets for me, this proposal has in its discussion phase treaded lightly on the fact that it effects others.

I would like to understand the rational given that the AMM was excluded from the scope of this? Yet other non-spendable transaction types like Escrows (for IOU's XLS-34) are not. Both are at similar stage of completed.

Possibly going forward, a review of what amendments are effected, proposed as well us the current working body of work should be one of the agreed upon gates before the XLS gets closed.

shawnxie999:
@lathanbritz Perhaps the spec was not clear. Payment channel and escrow of IOU is excluded from as well. In other words, clawback cannot clawback the tokens that are locked up in escrow or payment channel. It would not affect other amendments

lathanbritz:
Thank you @shawnxie999 maybe the review process could specifically require a effected/not-effected table as the XLS agreed upon and closed, would be very helpful.

shawnxie999:
>maybe the review process could specifically require a effected/not-effected table as the XLS agreed upon and closed, would be very helpful.

I think that's a good idea. Should we update the template to introduce a new section for this? @intelliot

PR #119, opened 2023-07-11
XLS-39d: rename `lsfAllowClawback` to `lsfAllowTrustLineClawback`

shawnxie999:
We are renaming `lsfAllowClawback` to `lsfAllowTrustLineClawback` to make it clearer that this flag is specifically used for IOUs. By renaming it, we want to separate this flag from potential future token types (like CFT) that may also support clawback functionality. This change is aimed at addressing technical debt caused by IOUs, so that we can avoid adding new flags at the account level in the future. Instead, we plan to control settings at the token level, rather than the account level. 

Eg. CFT will have a token-level toggle for clawback and **not** use `lsfAllowClawback`, so it's better to rename this flag to be explicit that it is only used for trust line, to avoid confusions in the future.

### DIFF
--- a/XLS-39d-clawback/README.md
+++ b/XLS-39d-clawback/README.md
@@ -8,7 +8,7 @@
 
 ## 1. Abstract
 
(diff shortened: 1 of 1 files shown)


kennyzlei (approved):
lgtm, thank you for the explanation of this rename

ckniffen:
Should `asfAllowClawback` be renamed as well?

ckniffen:
I think it should be mentioned in the spec as that is how you enable and asf have different values than lsf

shawnxie999:
@ckniffen I don't think other specs mention it either. I guess it's implied that they will have the same name 🤷

sappenin (approved):

PR #120, opened 2023-07-12
add details about clawback to XLS-33

ledhed2222:
Title says it all. LMK if you prefer I batch changes and keep a draft PR open instead.


Silkjaer (approved):

Silkjaer:
Whatever makes sense. If there's a lot at the same time, I think batch changes are preferable :)

PR #131, opened 2023-08-23
Change `cfts_by_issuer` API

shawnxie999:
- Fixed `ledger_index` to make it show the last modified ledger 
- Deleted `CFTokenIssuances` no longer need to display full data info, since it can be cumbersome to implement and not much incentive as users probably don't care about the info for deleted CFTs. If they really want to see deleted CFT info, they can use `deleted_ledger_index` to query for the state of CFT at the time of deletion, using `ledger_entry` API.


ledhed2222:
I don't understand why `ledger_index` is removed.

ledhed2222 (approved):

shawnxie999:
@Silkjaer Could this get merged ? thanks!

PR #135, opened 2023-09-28
Allow multiple bridges on a door account

seelabs:
The old spec required one bridge per door account. There was a request to allow multiple bridges on a door account. This patch updates the spec to allow this (with certain constraints).


Silkjaer (approved):

XLS-0038d-cross-chain-bridge/README.md
```diff
@@ -301,7 +301,13 @@ See the `XChainAddAccountCreateAttestation` section for more details on what thi
...
+The transaction must be submitted by the door account. It must also be submitted on both chains that form the bridge in order to be a valid bridge. A door account may own multiple bridge objects, subject to a few constraints.
+
+A bridge object must be owned by either the locking door account or the issuing chain door account. It is an error to try to create both sides of the bridge on the same chain. Doing so will result in a `tecDUPLICATE` error. For example, for a bridge with a locking door account of "Alice" and and issuing door account of "Bob" that locks "USD/gw" and issues "USD/bob". If Alice has already created this bridge, and Bob also tries to create this bridge _on the same chain as Alice_ then Bob's transaction will fail with a `tecDUPLICATE` error. It might appear that this constraint allows an account to try to block bridge creation by creating the other side of bridge accounts themselves. However, this would require them to either have the secret key for the door account or allow them to replay transactions (sidechains should have network ids that prevent transaction replay).
```
mvadari, line 306:
```suggestion
A bridge object must be owned by either the locking door account or the issuing chain door account. It is an error to try to create both sides of the bridge on the same chain. Doing so will result in a `tecDUPLICATE` error. For example, for a bridge with a locking door account of "Alice" and an issuing door account of "Bob" that locks "USD/gw" and issues "USD/bob", if Alice has already created this bridge, and Bob also tries to create this bridge _on the same chain as Alice_, then Bob's transaction will fail with a `tecDUPLICATE` error. It might appear that this constraint allows an account to try to block bridge creation by creating the other side of bridge accounts themselves. However, this would require them to either have the secret key for the door account or allow them to replay transactions (sidechains should have network ids that prevent transaction replay).
```

mvadari (approved):

PR #138, opened 2023-10-06
XLS-47d: Price Oracle

gregtatcam:
Discussions thread can be found here: https://github.com/XRPLF/XRPL-Standards/discussions/129


gregtatcam:
@aanchal4 and @mDuo13 please review.

gregtatcam:
Just FYI to the specs reviewers - unless something comes up, the implementation PR is expected to be opened soon - either by the end of this week (10/26) or early next week (10/30).

XLS-47d-PriceOracles/README.md
```diff
@@ -0,0 +1,453 @@
...
+
+  - `Symbol` is the symbol to be priced. Any arbitrary value should be allowed and interpreted exactly like other asset code fields in the ledger. A new enum value STI_CURRENCY and class STCurrency are introduced to support the `CURRENCY` field.
+  - `PriceUnit` is the denomination in which the prices are expressed. Any arbitrary value should be allowed and interpreted exactly like other asset code fields in the ledger.
+  - `SymbolPrice` is the scaled asset price, which is the price value after applying the scaling factor. This is an optional field. It is not included if the last update transaction didn't include the `Symbol`/`PriceUnit` pair.
+  - `Scale` is the price's scaling factor. It represents the price's precision level. For instance, if `Scale` is `6` and the original price is `0.155` then the scaled price is `155000`. Formally, $scaledPrice = originalPrice*{10}^{scale}$. Valid `Scale` range is {0-10}. This is an optional field. It is not included if the last update transaction didn't include the `Symbol`/`PriceUnit` pair.
+
+- `URI` is an optional URI field to reference the price data off-chain. It is a string of up to 256 ASCII hex encoded characters (0x20-0x7E).
```
sappenin, line 57: It's probably enough to just say this is a URI, correct? We can let that spec (e.g., [RFC-3986](https://datatracker.ietf.org/doc/html/rfc3986)) define the types of allowed characters.

If you agree, we could change this to just:

```suggestion
- `URI` is an optional [URI](https://datatracker.ietf.org/doc/html/rfc3986) field to reference price data off-chain. It is limited to 256 bytes.
```

sappenin (changes requested):

XLS-47d-PriceOracles/README.md
```diff
@@ -0,0 +1,453 @@
...
+
+### Terminology
+
+- Oracle Provider is a service or technology that enables the integration of external data and real-world events into a blockchain network.
+- dApp, short for decentralized application, refers to an application that is built on a blockchain network and operates using smart contracts or rely on other mechanisms or protocols for their functionality.
```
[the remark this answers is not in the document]
mvadari:
```suggestion
- **dApp (Decentralized Application)**: An application that is built on a blockchain network and operates using smart contracts or other on-chain mechanisms or protocols for their functionality.
```

XLS-47d-PriceOracles/README.md
```diff
@@ -0,0 +1,453 @@
...
+| `Provider` | :heavy_check_mark: | `string` | `BLOB` |
+| `PriceDataSeries` | :heavy_check_mark: | `array` | `ARRAY` |
+| `LastUpdateTime` | :heavy_check_mark: | `number` | `UINT32` |
+| `URI` | | `string` | `BLOB` |
+| `SymbolClass` | :heavy_check_mark: | `string` | `BLOB` |
+| `PreviousTxnID` | :heavy_check_mark: | `string` | `UINT256` |
+| `PreviousTxnLgrSeq`| :heavy_check_mark: | `string` | `UINT32` |
+
+- `LedgerEntryType` identifies the type of ledger object. The proposal recommends the value 0x0080 as the reserved entry type.
+- `Owner` is the account that has the update and delete privileges. It is recommended that this account has an associated [`signer list`](https://xrpl.org/set-up-multi-signing.html).
```
mvadari, line 41:
```suggestion
- `Owner` is the account that owns this object and has the update and delete privileges. It is recommended that this account has an associated [`signer list`](https://xrpl.org/set-up-multi-signing.html).
```

XLS-47d-PriceOracles/README.md
```diff
@@ -0,0 +1,453 @@
...
+- `SymbolClass` describes a type of the assets, for instance "currency", "commodity", "index". It is a string of up to ten ASCII hex encoded characters (0x20-0x7E).
+- `LastUpdateTime` is the specific point in time when the data was last updated. The `LastUpdateTime` is the ripple epoch time.
+- `PreviousTxnID` is the hash of the previous transaction to modify this entry. (Same as on other objects with this field.).
+- `PreviousTxnLgrSeq` is the ledger index of the ledger when this object was most recently updated/created. (Same as other objects with this field.)
+
+##### The `PriceOracle` Object ID Format
+
+We compute the `PriceOracle` object ID, a.k.a., `OracleID`, as the SHA-512Half of the following values, concatenated in order:
+
+- The Oracle space key (0x52)
```
mvadari, line 67: Is there a problem with having a difference in space key and ledger entry type? Seems like the last few amendments have had the same values for the two.

XLS-47d-PriceOracles/README.md
```diff
@@ -0,0 +1,453 @@
...
+- The Owner Account ID
+- The Oracle Sequence. This field must be passed to the transactions and it describes a unique Price Oracle sequence for the given account.
```
mvadari, line 69: Is this the transaction `Sequence` or a different number?

XLS-47d-PriceOracles/README.md
```diff
@@ -0,0 +1,453 @@
...
+- `PriceDataSeries` is an array of up to ten `PriceData` objects, where `PriceData` represents the price information for a token pair. `PriceOracle` with more than five `PriceData` objects requires two owner reserves. `PriceData` includes the following fields:
+
+  |FieldName | Required? | JSON Type | Internal Type|
+  |:---------|:-----------|:---------------|:---------------|
+  | `Symbol` | :heavy_check_mark: | `string` | `CURRENCY` |
+  | `PriceUnit` | :heavy_check_mark: | `string` | `CURRENCY` |
+  | `SymbolPrice` | | `number` | `UINT64` |
+  | `Scale` | | `number` | `UINT8` |
+
+  - `Symbol` is the symbol to be priced. Any arbitrary value should be allowed and interpreted exactly like other asset code fields in the ledger. A new enum value STI_CURRENCY and class STCurrency are introduced to support the `CURRENCY` field.
```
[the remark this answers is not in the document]
mvadari: ++ - I was a bit confused about the difference between `Symbol` and `PriceUnit`

XLS-47d-PriceOracles/README.md
```diff
@@ -0,0 +1,453 @@
...
+- The Oracle Provider has to publish the `Account` account public key so that it can be used for verification by dApp’s.
+- The Oracle Provider has to publish a registry of available Price Oracles with their unique Oracle Sequence. The hash of the `Account` and the `OracleSequence` uniquely identifies the Price Oracle on-ledger object.
+
+#### Transaction fields for **SetOracle** transaction
+
+|FieldName | Required? | JSON Type | Internal Type |
+|:---------|:-----------|:---------------|:------------|
+| `TransactionType` | :heavy_check_mark: | `string`| UINT16 |
+
+Indicates a new transaction type `SetOracle`. The integer value is 49.
```
mvadari, line 112: The transaction numbers need to be updated after DID. I recommend leaving them off the spec.

mvadari, line 112: Shouldn't the transactions be called `OracleSet` and `OracleDelete` to match current conventions?

XLS-47d-PriceOracles/README.md
```diff
@@ -0,0 +1,453 @@
...
+
+- `LedgerEntryType` identifies the type of ledger object. The proposal recommends the value 0x0080 as the reserved entry type.
+- `Owner` is the account that has the update and delete privileges. It is recommended that this account has an associated [`signer list`](https://xrpl.org/set-up-multi-signing.html).
+- `Provider` identifies an Oracle Provider. It can be URI or any data, for instance `chainlink`. It is a string of up to 256 ASCII hex encoded characters (0x20-0x7E).
+- `PriceDataSeries` is an array of up to ten `PriceData` objects, where `PriceData` represents the price information for a token pair. `PriceOracle` with more than five `PriceData` objects requires two owner reserves. `PriceData` includes the following fields:
+
+  |FieldName | Required? | JSON Type | Internal Type|
+  |:---------|:-----------|:---------------|:---------------|
+  | `Symbol` | :heavy_check_mark: | `string` | `CURRENCY` |
+  | `PriceUnit` | :heavy_check_mark: | `string` | `CURRENCY` |
```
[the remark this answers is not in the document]
gregtatcam: Say in a `BTC/USD` trading pair, `BTC` is the `Symbol` and `USD` is the `PriceUnit`. For instance a price of 36,180 for a `BTC/USD` means that one unit of `BTC` valued at 36,180 units of `USD`. How about changing these two fields to the following to provide more clarity:

- `BaseAsset`: This refers to the primary asset within a trading pair. It is the asset against which the price of the quote asset is quoted. The base asset is usually considered the 'primary' asset and forms the basis for trading. For example:
In the pair `BTC/USD`, `BTC` is the base asset.
In `ETH/BTC`, `ETH` is the base asset.

- `QuoteAsset`: This represents the secondary or quote asset in a trading pair. It denotes the price of one unit of the base asset. The quote asset's value is expressed in terms of the base asset. For example:
In the pair `BTC/USD`, `USD` is the quote asset.
In `ETH/BTC`, `BTC` is the quote asset.
The base asset and quote asset together form a trading pair, and their relationship determines the price at which one asset can be exchanged for another.

The `CURRENCY` can be changed to `ASSET_TYPE` to give a more generalized meaning. The `ASSET_TYPE` is either the three-letter currency or some other asset represented as a hexadecimal string. Either way it's a 160-bit field, which is conformant to the XRPL currency description: https://xrpl.org/currency-formats.html#currency-codes.

There is also a `SymbolClass` field, which provides additional information for the `BaseAsset`. I'm proposing to change it to `AssetClass` to reflect the above updates.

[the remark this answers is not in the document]
sappenin: Yes and yes. Agree with everything - what you wrote above is much clearer!

XLS-47d-PriceOracles/README.md
```diff
@@ -0,0 +1,453 @@
...
+- The `Account` account doesn't exist or the `Account` is not equal to the `Owner` field when updating the Oracle instance.
+- The transaction is not signed by the `Account` account or the account's multi signers.
+- The `URI` field length exceeds 64 bytes.
+- The `Provider` field length exceeds 64 bytes.
+- The `SymbolClass` field length exceeds 12 bytes.
+- The `LastUpdateTime` field is less than the previous `LastUpdateTime` or is greater than the last close time plus 30 seconds.
+
+If an object with the `OracleID` Object ID already exists then the new token pairs are added to the Oracle instance `PriceDataSeries` array. Note that the order of the token pairs in the `PriceDataSeries` array is not important since the token pair uniquely identifies location in the `PriceDataSeries` array of the `PriceOracle` object. Also note that not every token pair price has to be updated. I.e., even though the `PriceOracle` may define ten token pairs, `SetOracle` transaction may contain only one token pair price update. In this case the missing token pair will not include `SymbolPrice` and `Scale` fields. `PreviousTxnID` can be used to find the last updated Price Data for this token pair.
```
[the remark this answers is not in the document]
gregtatcam: Currently there is no provision for this. There are several arguments that could be made against the deletion:
- The addition of deletion functionality increases the complexity of maintaining and documenting the Price Oracle.
- Deletion would introduce complexity for decentralized applications (dApps) reliant on the Price Oracle.
- The need for deletion suggests the requirement for pair updates, which could introduce additional complexity.
- Pair deletions may interfere with viewing historical data associated with removed pairs. 
- Frequent deletion and potential recreation could impact the overall reliability of the Price Oracle.

XLS-47d-PriceOracles/README.md
```diff
@@ -0,0 +1,453 @@
...
+  | `PriceUnit` | :heavy_check_mark: | `string` | `CURRENCY` |
+  | `SymbolPrice` | | `number` | `UINT64` |
+  | `Scale` | | `number` | `UINT8` |
+
+  - `Symbol` is the symbol to be priced. Any arbitrary value should be allowed and interpreted exactly like other asset code fields in the ledger. A new enum value STI_CURRENCY and class STCurrency are introduced to support the `CURRENCY` field.
+  - `PriceUnit` is the denomination in which the prices are expressed. Any arbitrary value should be allowed and interpreted exactly like other asset code fields in the ledger.
+  - `SymbolPrice` is the scaled asset price, which is the price value after applying the scaling factor. This is an optional field. It is not included if the last update transaction didn't include the `Symbol`/`PriceUnit` pair.
+  - `Scale` is the price's scaling factor. It represents the price's precision level. For instance, if `Scale` is `6` and the original price is `0.155` then the scaled price is `155000`. Formally, $scaledPrice = originalPrice*{10}^{scale}$. Valid `Scale` range is {0-10}. This is an optional field. It is not included if the last update transaction didn't include the `Symbol`/`PriceUnit` pair.
```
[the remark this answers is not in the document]
gregtatcam: It is a pretty long description for a fairly straightforward field. And the example/formula provide an adequate clarity. I think I'm going to keep the original one.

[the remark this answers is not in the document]
sappenin: Sounds good - I'll resolve this one.

XLS-47d-PriceOracles/README.md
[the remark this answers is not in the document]
gregtatcam: @sappenin I think we should keep the `CURRENCY` because the internal type is `Currency` and it can't be changed unless a duplicate type `AssetType` is introduced behaving exactly like the `Currency`. This doesn't make much sense. Another option is use the `VL` (blob) type of say up to 20 bytes. The goodness of the `Currency` is the standard currency (3-letters) can be passed as a regular string but the drawback is that anything else has to be a 160-bit hexadecimal, so we are going to waste some space. The advantage of the `VL` is that it's a variable length so no space is wasted but the drawback is that all input has to be passed as a hexadecimal.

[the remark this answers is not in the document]
sappenin: +1 on keeping `CURRENCY`. 

Unless I missed anything, I think this comment can be resolved @gregtatcam once you've captured everything useful from it.

XLS-47d-PriceOracles/README.md
```diff
@@ -0,0 +1,468 @@
...
+   Affiliation:  <a href="https://ripple.com">Ripple</a>
+ 
+ </pre>
+ 
+ # Price Oracles on XRP Ledger
+ 
+ ## Abstract
+
+ This proposal adds an on-chain `PriceOracle` object to the XRPL ledger. A blockchain oracle is a system or service that acts as a bridge between a blockchain network and the external world, providing off-chain data or information to decentralized applications (dApps) on the blockchain. Oracles are used to bring real-world data, for instance  market prices, exchange rates, interest rates, or weather conditions onto the blockchain, enabling dApps to access and utilize information that resides outside the blockchain. This document outlines a new protocol for price oracles the on XRPL ledger, and provides guidelines for developers and system architects to implement and utilize this solution effectively. This proposal introduces a new on-ledger `PriceOracle` object and the transactions to create, delete, and update the `PriceOracle`. It also adds the `get_aggregate_price` API, to retrieve an `aggregate mean`, `trimmed mean`, and `median` for the provided price oracles. This feature requires an amendment.
```
sappenin, line 13: Just clarifying -- is the intention that this spec should support non-price data like Weather data, or is that just an example of a more general oracle?

gregtatcam: The intention is to support Price Data only.

sappenin: Awesome, thanks for clarifying and updating the title.

XLS-47d-PriceOracles/README.md
```diff
@@ -0,0 +1,468 @@
...
+      },
+      "validated" : false
+      "time" : 78937648
+    }
+
+## Appendices
+
+### Appendix 1. STI_CURRENCY
+
+A new type, `STI_CURRENCY`, is introduced to support `BaseAsset` and `QuoteAsset` fields' type `CURRENCY`. This type can represent a standard currency code or an arbitrary asset as a 160-bit (40 character) hexadecimal string. This type is conformant to the XRPL [Currency Codes](https://xrpl.org/currency-formats.html#currency-codes). Below is a JSON example with the `BaseAsset` representing a CUSIP code `912810RR9` as a 160-bit hexadecimal string and a `QuoteAsset` representing a standard `USD` currency code:
```
Bronek, line 459: https://xrpl.org/currency-formats.html#currency-codes explicitly disallows XRP, it would be good to mention that in this context XRP is allowed.

XLS-47d-PriceOracles/README.md
```diff
@@ -0,0 +1,468 @@
...
+
+|FieldName | Required? | JSON Type |
+|:---------|:-----------|:---------------|
+| `time_threshold` | | `number` |
+
+The `time_threshold` is used to define a time range in seconds for filtering out older price data. It's an optional parameter and is 0 by default; i.e. there is no filtering in this case.
+
+The price data to aggregate is selected based on specific criteria. The most recent Price Oracle object is obtained for the specified oracles. The most recent `LastUpdateTime` among all objects is chosen as the upper time threshold. A Price Oracle object is included in the aggregation dataset if it satisfies the conditions of containing the specified `base_asset`/`quote_asset` pair, including the `AssetPrice` field, and its `LastUpdateTime` is within the time range of (upper threshold - time threshold) to the upper threshold. If a Price Oracle object doesn't contain the `AssetPrice` for the specified token pair, then up to three previous Price Oracle objects are examined and include the first one that fulfills the criteria.
```
Bronek, line 373: "first one that fulfills" could be confusing. I would use "most recent one" here

XLS-47d-PriceOracles/README.md
```diff
@@ -0,0 +1,468 @@
...
+###### Example of get_aggregate_price API JSON request
+
+    {
+    "method": "get_aggregate_price",
+    "params": [
+        {
+            "ledger_index": "current",
+            "base_asset": "XRP",
+            "quote_asset": "USD",
+            "flags": 7,
```
Bronek, line 392: what this ? Cannot see `flags` defined above.

XLS-47d-PriceOracles/README.md
```diff
@@ -0,0 +1,468 @@
...
+  |:---------|:-----------|:---------------|
+  | `oracle_sequence` | :heavy_check_mark: | number |
+
+  `oracle_sequence` is a unique identifier of the Price Oracle for the given Account.
+
+|FieldName | Required? | JSON Type |
+|:---------|:-----------|:---------------|
+| `trim` | | `number` |
+
+`trim` is the percentage of outliers to trim. Valid trim range is 1-25. If this parameter is included then the API returns trimmed median in addition to the average and median.
```
Bronek, line 365: "trimmed median" ? I think the term is "trimmed mean". Also, would suggest using "mean" instead of average through because it is more precise.

XLS-47d-PriceOracles/README.md
gregtatcam: Thanks. It's a leftover from a first version, which did have the flags. Removed.

XLS-47d-PriceOracles/README.md
gregtatcam: Thanks. It's a typo. The implementation has the trimmed average. I'm ok with using `mean` instead.

XLS-47d-PriceOracles/README.md
```diff
@@ -0,0 +1,468 @@
...
+
+The `PriceOracle` ledger entry represents the `PriceOracle` object on XRPL ledger and contains the following fields:
+
+|FieldName | Required? | JSON Type | Internal Type|
+|:---------|:-----------|:---------------|:---------------|
+| `LedgerEntryType`| :heavy_check_mark: | `string` | `UINT16` |
+| `Owner` | :heavy_check_mark: | `string` | `ACCOUNTID` |
+| `Provider` | :heavy_check_mark: | `string` | `BLOB` |
+| `PriceDataSeries` | :heavy_check_mark: | `array` | `ARRAY` |
+| `LastUpdateTime` | :heavy_check_mark: | `number` | `UINT32` |
```
Bronek, line 34: how about using string as ISO8601-compliant time, with second precision ? I heard that some users have problems using ripple-epoch time.

XLS-47d-PriceOracles/README.md
```diff
@@ -0,0 +1,468 @@
...
+|:---------|:-----------|:---------------|:------------|
+| `AssetClass` | | `string` | `BLOB` |
+
+`AssetClass` describes the asset's type.
+
+|FieldName | Required? | JSON Type | Internal Type |
+|:---------|:-----------|:---------------|:-----------|
+| `LastUpdateTime` | :heavy_check_mark: | `number` | `UINT32` |
+
+`LastUpdateTime` is the specific point in time when the data was last updated.
```
Bronek, line 148: same comment as above - how about using ISO8601-compliant string ?

XLS-47d-PriceOracles/README.md
```diff
@@ -0,0 +1,468 @@
...
+- `entire_set` is an object of the following fields:
+  - `size` is the size of the data set used to calculate the statistics.
+  - `average` is the simple average.
+  - `standard_deviation` is the standard deviation.
+- `trimmed_set` is an object, which is included in the response if `trim` fields is set. The object has the following fields:
+  - `size` is the size of the data set used to calculate the statistics.
+  - `average` is the simple average.
+  - `standard_deviation` is the standard deviation.
+- `median` is the median.
+- `time` is the most recent time stamp out of all `LastUpdateTime` values.
```
Bronek, line 433: if my proposal of using ISO8601 time, then this line does not need to change, but the examples below do.

XLS-47d-PriceOracles/README.md
Bronek: FWIW I do not feel strongly about ISO8601, but it would lower adoption bar a little.

XLS-47d-PriceOracles/README.md
gregtatcam: Thanks for bringing this up. After some discussion decided to change to Unix Time.

gregtatcam: This is a comment from @thejohnfreeman : https://github.com/XRPLF/rippled/pull/4789#issuecomment-1819874166

I find the [description](https://github.com/XRPLF/XRPL-Standards/blob/e0a27d1e09234f422db83420ff7dcd62d3a54edc/XLS-47d-PriceOracles/README.md#example-of-priceoracle-json) of the `OracleID` confusing. The spec says this:

> We compute the PriceOracle object ID, a.k.a., OracleID, as the SHA-512Half of the following values, concatenated in order:
>  - The Oracle space key (0x52)
>  - The Owner Account ID
>  - The Oracle Sequence. This field must be passed to the transactions and it describes a unique Price Oracle sequence for the given account.

- `OracleID` is not mentioned in the [description](https://github.com/XRPLF/XRPL-Standards/blob/e0a27d1e09234f422db83420ff7dcd62d3a54edc/XLS-47d-PriceOracles/README.md#the-priceoracle-object) of the fields of `PriceOracle`, but it is [included](https://github.com/XRPLF/XRPL-Standards/blob/e0a27d1e09234f422db83420ff7dcd62d3a54edc/XLS-47d-PriceOracles/README.md#example-of-priceoracle-json) in the JSON of an example `PriceOracle`. Is it included in the ledger object or not?
- It is implied but not explicitly stated that the `OracleID` is used as the `SHAMap` key for a `PriceOracle` object. Is it? How are these objects supposed to be laid out in the ledger? The aggregation transaction suggests that it is possible to iterate `PriceOracle` objects for a given account. How?
- Oracle Sequence is not mentioned in the fields of `PriceOracle` or shown in the example. Is it possible to recover the `OracleID` for a `PriceOracle` using only the fields contained within a `PriceOracle`? If so, how?
- "Sequence" implies to me that it is a number that increments by one with each change, like a Ledger Sequence or an Account Sequence, but `OracleID` implies a constant identifier. Which is it? Are there any rules on the sequence number?

In the spec, the [list](https://github.com/XRPLF/XRPL-Standards/blob/e0a27d1e09234f422db83420ff7dcd62d3a54edc/XLS-47d-PriceOracles/README.md#user-content-transaction-fields-for-oracleset-transaction) following "The transaction fails if" includes:

> - The `URI` field length exceeds 64 bytes.
> - The `Provider` field length exceeds 64 bytes. 

But the [section](https://github.com/XRPLF/XRPL-Standards/blob/e0a27d1e09234f422db83420ff7dcd62d3a54edc/XLS-47d-PriceOracles/README.md#user-content-the-priceoracle-object) describing the `PriceOracle` on-ledger object fields says:

> - `Provider` identifies an Oracle Provider. It can be URI or any data, for instance chainlink. It is a string of up to 256 ASCII hex encoded characters (0x20-0x7E).
> - `URI` is an optional [URI](https://datatracker.ietf.org/doc/html/rfc3986) field to reference price data off-chain. It is limited to 256 bytes.

Which is correct? Can we fix it in the spec?

I find this explanation unclear:

> If an object with the `OracleID` Object ID already exists then the new token pairs are added to the Oracle instance `PriceDataSeries` array. Note that the order of the token pairs in the `PriceDataSeries` array is not important since the token pair uniquely identifies location in the `PriceDataSeries` array of the `PriceOracle` object. Also note that not every token pair price has to be updated. I.e., even though the `PriceOracle` may define ten token pairs, `OracleSet` transaction may contain only one token pair price update. In this case the missing token pair will not include `AssetPrice` and `Scale` fields. `PreviousTxnID` can be used to find the last updated Price Data for this token pair.

"If an object with the ID exists, then the new token pairs are added..." I'm assuming "new token pairs" are ones that do not appear in the `PriceOracle`. If so, then why no mention of "old token pairs"? They should be updating what appears in the `PriceOracle`, correct?

"In this case the missing token pair..." What missing token pair? The token pairs that are in the object, but not the transaction? If yes, then what is meant by "[they] will not include `AssetPrice` and `Scale` fields"? If they are missing, won't they "not include" _every_ field?

I feel like there is an implied behavior for this transaction that is not well described by the existing language. Here is how I would describe that behavior. Is it correct?

> An `OracleSet` transaction uniquely identifies a `PriceOracle` object with its `Account` and `OracleSequence` fields. If such an object does not yet exist in the ledger, it is created. Otherwise, the existing object is updated. The `Provider`, `URI`, and `AssetClass` fields are copied directly from the transaction, if present. `Provider` and `AssetClass` must be included in the transaction if the object is being created.
>
> The `PriceDataSeries` of the transaction is copied to a newly created `PriceOracle` object, or updates an existing object, like so:
> - `PriceData` objects for (`BaseAsset`, `QuoteAsset`) token pairs that appear in the transaction but not the object are copied to the object.
> - `PriceData` objects for token pairs that appear in both the transaction and the object are overwritten in the object.
> - `PriceData` objects for token pairs that appear only in the object are left unchanged.
>
> The order of token pairs in the transaction is not important because the token pair uniquely identifies the location of the `PriceData` object in the `PriceDataSeries` array of the `PriceOracle` object.
>
> `LastUpdateTime`, `PreviousTxnID`, and `PreviousTxnLgrSeq` are set in the same manner as for an `AccountSet` transaction.
>
> The owner reserve of the account is updated according to the difference in the size of the `PriceDataSeries` before and after the transaction is applied: 0 for missing, 1 for 1 - 5 objects, 2 for 6 - 10 objects.

`AssetClass` is a required field on a `PriceOracle` object and an optional field in an `OracleSet` transaction, just like `Provider`, but its field description does not say that it must be included in the transaction when creating a new instance of `PriceOracle`, unlike `Provider`. It must be included in the transaction, correct?

A side note on formatting. I think the presentation in the section on `PriceOracle` fields is best: _one_ table with all the fields, followed by a bulleted list of field descriptions. Not like the transaction sections with separate tables (repeating the same header row) for each field. I think the example should go directly after or before the table, not at the end of the section.

gregtatcam:
This is a comment from @thejohnfreeman : https://github.com/XRPLF/rippled/pull/4789#issuecomment-1819874166

I find the [description](https://github.com/XRPLF/XRPL-Standards/blob/e0a27d1e09234f422db83420ff7dcd62d3a54edc/XLS-47d-PriceOracles/README.md#example-of-priceoracle-json) of the `OracleID` confusing. The spec says this:

> We compute the PriceOracle object ID, a.k.a., OracleID, as the SHA-512Half of the following values, concatenated in order:
>  - The Oracle space key (0x52)
>  - The Owner Account ID
>  - The Oracle Sequence. This field must be passed to the transactions and it describes a unique Price Oracle sequence for the given account.

- `OracleID` is not mentioned in the [description](https://github.com/XRPLF/XRPL-Standards/blob/e0a27d1e09234f422db83420ff7dcd62d3a54edc/XLS-47d-PriceOracles/README.md#the-priceoracle-object) of the fields of `PriceOracle`, but it is [included](https://github.com/XRPLF/XRPL-Standards/blob/e0a27d1e09234f422db83420ff7dcd62d3a54edc/XLS-47d-PriceOracles/README.md#example-of-priceoracle-json) in the JSON of an example `PriceOracle`. Is it included in the ledger object or not?
- It is implied but not explicitly stated that the `OracleID` is used as the `SHAMap` key for a `PriceOracle` object. Is it? How are these objects supposed to be laid out in the ledger? The aggregation transaction suggests that it is possible to iterate `PriceOracle` objects for a given account. How?
- Oracle Sequence is not mentioned in the fields of `PriceOracle` or shown in the example. Is it possible to recover the `OracleID` for a `PriceOracle` using only the fields contained within a `PriceOracle`? If so, how?
- "Sequence" implies to me that it is a number that increments by one with each change, like a Ledger Sequence or an Account Sequence, but `OracleID` implies a constant identifier. Which is it? Are there any rules on the sequence number?

In the spec, the [list](https://github.com/XRPLF/XRPL-Standards/blob/e0a27d1e09234f422db83420ff7dcd62d3a54edc/XLS-47d-PriceOracles/README.md#user-content-transaction-fields-for-oracleset-transaction) following "The transaction fails if" includes:

> - The `URI` field length exceeds 64 bytes.
> - The `Provider` field length exceeds 64 bytes. 

But the [section](https://github.com/XRPLF/XRPL-Standards/blob/e0a27d1e09234f422db83420ff7dcd62d3a54edc/XLS-47d-PriceOracles/README.md#user-content-the-priceoracle-object) describing the `PriceOracle` on-ledger object fields says:

> - `Provider` identifies an Oracle Provider. It can be URI or any data, for instance chainlink. It is a string of up to 256 ASCII hex encoded characters (0x20-0x7E).
> - `URI` is an optional [URI](https://datatracker.ietf.org/doc/html/rfc3986) field to reference price data off-chain. It is limited to 256 bytes.

Which is correct? Can we fix it in the spec?

I find this explanation unclear:

> If an object with the `OracleID` Object ID already exists then the new token pairs are added to the Oracle instance `PriceDataSeries` array. Note that the order of the token pairs in the `PriceDataSeries` array is not important since the token pair uniquely identifies location in the `PriceDataSeries` array of the `PriceOracle` object. Also note that not every token pair price has to be updated. I.e., even though the `PriceOracle` may define ten token pairs, `OracleSet` transaction may contain only one token pair price update. In this case the missing token pair will not include `AssetPrice` and `Scale` fields. `PreviousTxnID` can be used to find the last updated Price Data for this token pair.

"If an object with the ID exists, then the new token pairs are added..." I'm assuming "new token pairs" are ones that do not appear in the `PriceOracle`. If so, then why no mention of "old token pairs"? They should be updating what appears in the `PriceOracle`, correct?

"In this case the missing token pair..." What missing token pair? The token pairs that are in the object, but not the transaction? If yes, then what is meant by "[they] will not include `AssetPrice` and `Scale` fields"? If they are missing, won't they "not include" _every_ field?

I feel like there is an implied behavior for this transaction that is not well described by the existing language. Here is how I would describe that behavior. Is it correct?

> An `OracleSet` transaction uniquely identifies a `PriceOracle` object with its `Account` and `OracleSequence` fields. If such an object does not yet exist in the ledger, it is created. Otherwise, the existing object is updated. The `Provider`, `URI`, and `AssetClass` fields are copied directly from the transaction, if present. `Provider` and `AssetClass` must be included in the transaction if the object is being created.
>
> The `PriceDataSeries` of the transaction is copied to a newly created `PriceOracle` object, or updates an existing object, like so:
> - `PriceData` objects for (`BaseAsset`, `QuoteAsset`) token pairs that appear in the transaction but not the object are copied to the object.
> - `PriceData` objects for token pairs that appear in both the transaction and the object are overwritten in the object.
> - `PriceData` objects for token pairs that appear only in the object are left unchanged.
>
> The order of token pairs in the transaction is not important because the token pair uniquely identifies the location of the `PriceData` object in the `PriceDataSeries` array of the `PriceOracle` object.
>
> `LastUpdateTime`, `PreviousTxnID`, and `PreviousTxnLgrSeq` are set in the same manner as for an `AccountSet` transaction.
>
> The owner reserve of the account is updated according to the difference in the size of the `PriceDataSeries` before and after the transaction is applied: 0 for missing, 1 for 1 - 5 objects, 2 for 6 - 10 objects.

`AssetClass` is a required field on a `PriceOracle` object and an optional field in an `OracleSet` transaction, just like `Provider`, but its field description does not say that it must be included in the transaction when creating a new instance of `PriceOracle`, unlike `Provider`. It must be included in the transaction, correct?

A side note on formatting. I think the presentation in the section on `PriceOracle` fields is best: _one_ table with all the fields, followed by a bulleted list of field descriptions. Not like the transaction sections with separate tables (repeating the same header row) for each field. I think the example should go directly after or before the table, not at the end of the section.

gregtatcam:
@thejohnfreeman. Thanks for the feedback.

I removed the references to the `OracleID`. It's not used anywhere. The `PriceOracle` object is identified by `Account` and `OracleDocumentID`. The latter replaces `OracleSequence`.

I corrected the validation fields for `URI` and `Provider` to 32 bytes.

I updated the `OracleSet` transaction description per your suggestion.

I followed the format of tables/description of other specs like AMM/NFT/DID. I agree that having one table followed by the description is better. I changed the format and moved the json examples for transactions and RPC before the tables.

I'll update the implementation If the updates look right to you.

XLS-47d-PriceOracles/README.md
```diff
@@ -0,0 +1,468 @@
...
+   Affiliation:  <a href="https://ripple.com">Ripple</a>
+ 
+ </pre>
+ 
+ # Price Oracles on XRP Ledger
+ 
+ ## Abstract
+
+ This proposal adds an on-chain `PriceOracle` object to the XRPL ledger. A blockchain oracle is a system or service that acts as a bridge between a blockchain network and the external world, providing off-chain data or information to decentralized applications (dApps) on the blockchain. Oracles are used to bring real-world data, for instance  market prices, exchange rates, interest rates, or weather conditions onto the blockchain, enabling dApps to access and utilize information that resides outside the blockchain. This document outlines a new protocol for price oracles the on XRPL ledger, and provides guidelines for developers and system architects to implement and utilize this solution effectively. This proposal introduces a new on-ledger `PriceOracle` object and the transactions to create, delete, and update the `PriceOracle`. It also adds the `get_aggregate_price` API, to retrieve an `aggregate mean`, `trimmed mean`, and `median` for the provided price oracles. This feature requires an amendment.
```
mvadari, line 13:
```suggestion
This proposal adds an on-chain `PriceOracle` object to the XRP Ledger. A blockchain oracle is a system or service that acts as a bridge between a blockchain network and the external world, providing off-chain data or information to decentralized applications (dApps) on the blockchain. Oracles are used to bring real-world data, for instance  market prices, exchange rates, interest rates, or weather conditions onto the blockchain, enabling dApps to access and utilize information that resides outside the blockchain. This document outlines a new protocol for price oracles the on XRPL ledger, and provides guidelines for developers and system architects to implement and utilize this solution effectively. This proposal introduces a new on-ledger `PriceOracle` object and the transactions to create, delete, and update the `PriceOracle`. It also adds the `get_aggregate_price` API, to retrieve an `aggregate mean`, `trimmed mean`, and `median` for the provided price oracles. This feature requires an amendment.
```

XLS-47d-PriceOracles/README.md
mvadari: WE already use ripple epoch time on-ledger in objects like escrows, I'd rather be internally consistent.

XLS-47d-PriceOracles/README.md
```diff
@@ -62,17 +62,18 @@ The `PriceOracle` ledger entry represents the `PriceOracle` object on XRPL ledge
 
 ##### The `PriceOracle` Object ID Format
 
-We compute the `PriceOracle` object ID, a.k.a., `OracleID`, as the SHA-512Half of the following values, concatenated in order:
+We compute the `PriceOracle` object ID as the SHA-512Half of the following values, concatenated in order:
 
 - The Oracle space key (0x52)
-- The Owner Account ID
-- The Oracle Sequence. This field must be passed to the transactions and it describes a unique Price Oracle sequence for the given account.
+- The Owner Account ID, `Account`.
```
thejohnfreeman, line 68: `Account` here but `Owner` in the next paragraph. I was guessing it should be `Owner` in both places (as it is in the fields table above and example below), but it is `Account` in the transactions. Can we unify? Should we? I'm guessing we use `Account` in the transactions because all transactions must have that field? Is that true?

XLS-47d-PriceOracles/README.md
```diff
@@ -186,82 +164,63 @@ The transaction fails if:
 - `PriceDataSeries` has array elements with missing `AssetPrice`.
 - The `Account` account doesn't exist or the `Account` is not equal to the `Owner` field when updating the Oracle instance.
 - The transaction is not signed by the `Account` account or the account's multi signers.
-- The `URI` field length exceeds 64 bytes.
-- The `Provider` field length exceeds 64 bytes.
+- The `URI` field length exceeds 32 bytes.
+- The `Provider` field length exceeds 32 bytes.
```
thejohnfreeman, line 168: There are mentions of 256 bytes in the fields list in the section titled "The `PriceOracle` Object". (I cannot make this comment on those lines in this PR because they were not changed.)

XLS-47d-PriceOracles/README.md
```diff
@@ -101,81 +102,58 @@ We define a new transaction **OracleSet** for creating or updating a `PriceOracl
...
-
-`AssetPrice` is the scaled asset price, which is the price value after applying the scaling factor.
-
-|FieldName | Required? | JSON Type | Internal Type |
-|:---------|:-----------|:---------------|:------------|
 | `Scale` | :heavy_check_mark: | `number` | `UINT8` |
 
-`Scale` is the price's scaling factor.
+- `TransactionType` Indicates a new transaction type `OracleSet`. The integer value is TBD.
+- `Account` is the XRPL account that has update and delete privileges on the Oracle being set.
```
thejohnfreeman, line 146: We should mention (here and elsewhere, in the other txn sections) that this field corresponds to the `Owner` field on the `PriceOracle` ledger object.

XLS-47d-PriceOracles/README.md
```diff
@@ -101,81 +102,58 @@ We define a new transaction **OracleSet** for creating or updating a `PriceOracl
...
-|FieldName | Required? | JSON Type | Internal Type |
-|:---------|:-----------|:---------------|:------------|
-| `OracleSequence` | :heavy_check_mark: | `string` | `UINT32` |
-
-`OracleSequence` is a unique identifier of the Price Oracle for the given Account.
-
-|FieldName | Required? | JSON Type | Internal Type |
-|:---------|:-----------|:---------------|:------------|
+| `OracleDocumentID` | :heavy_check_mark: | `string` | `UINT32` |
 | `Provider` | | `string` | `BLOB` |
```
thejohnfreeman, line 135: What do you think of a third marker (after "no marker" and :heavy_check_mark:) to indicate a field is conditionally required, like `Provider` and `AssetClass`? Perhaps :grey_question: ?

thejohnfreeman (commented):
Thanks for the changes. I have a few followup comments here, but I think I understand the spec well enough now to review the code.

XLS-47d-PriceOracles/README.md
gregtatcam: Correct, `Account` is used in the transactions. I'll update to `Owner` outside of the transactions.

mvadari (commented):
I don't think the `.DS_Store` file should be included in this PR (perhaps it needs to be added to `.gitignore`?

gregtatcam:
> I don't think the `.DS_Store` file should be included in this PR (perhaps it needs to be added to `.gitignore`?

Thanks. I must have selected all Unversioned files in CLion. Will delete.

thejohnfreeman (approved):
Looks great. Minor nits follow:

- "XRP ledger" -> "XRP Ledger"?
- "Unix Time stamp" -> "Unix timestamp"? Or "Ripple Epoch timestamp"? Or even just "timestamp"?

XLS-47d-PriceOracles/README.md
```diff
@@ -0,0 +1,403 @@
...
+
+- The Oracle space key (0x52)
+- The Owner Account ID, `Owner`.
+- The Oracle Document ID, `OracleDocumentID`. This field describes a unique Price Oracle instance for the given account. The Oracle Document ID is maintained by the Oracle Provider.
+
+The `Owner` and `OracleDocumentID` uniquely identify the `PriceOracle` object and must be passed to the Oracle transactions.
+
+#### Example of `PriceOracle` JSON
+
+    {
```
mvadari, line 75: nit: for ease of reading, perhaps the hex-encoded fields should have comments as to what their decoded values are.

XLS-47d-PriceOracles/README.md
```diff
@@ -0,0 +1,403 @@
...
+    }
+
+#### Transaction fields for **OracleSet** transaction
+
+|FieldName | Required? | JSON Type | Internal Type |
+|:---------|:-----------|:---------------|:------------|
+| `TransactionType` | :heavy_check_mark: | `string`| UINT16 |
+| `Account` | :heavy_check_mark: | `string` | `ACCOUNTID` |
+| `OracleDocumentID` | :heavy_check_mark: | `string` | `UINT32` |
+| `Provider` | :grey_question: | `string` | `BLOB` |
```
mvadari, line 135: The question mark here and in `AssetClass` make it seem like the spec is unfinished - IMO it should just be labeled as "not required" and further explained below.

XLS-47d-PriceOracles/README.md
```diff
@@ -0,0 +1,403 @@
...
+| `URI` | | `string` | `BLOB` |
+| `AssetClass` | :grey_question: | `string` | `BLOB` |
+| `LastUpdateTime` | :heavy_check_mark: | `number` | `UINT32` |
+| `PriceDataSeries` | :heavy_check_mark: | `array` | `ARRAY` |
+| `BaseAsset` | :heavy_check_mark: | `string` | `CURRENCY` |
+| `QuoteAsset` | :heavy_check_mark: | `string` | `CURRENCY` |
+| `AssetPrice` | :heavy_check_mark: | `number` | `UINT64` |
+| `Scale` | :heavy_check_mark: | `number` | `UINT8` |
+
+- `TransactionType` Indicates a new transaction type `OracleSet`. The integer value is TBD.
```
mvadari, line 145:
```suggestion
- `TransactionType` Indicates a new transaction type `OracleSet`.
```
IMO the integer value isn't necessary for the spec.

XLS-47d-PriceOracles/README.md
```diff
@@ -0,0 +1,403 @@
...
+
+#### Transaction fields for **OracleDelete** transaction
+
+|FieldName | Required? | JSON Type |
+|:---------|:-----------|:---------------|
+| `TransactionType` | :heavy_check_mark: | `string` | `UINT16` |
+| `Account` | :heavy_check_mark: | `string` | `ACCOUNTID` |
+| `OracleDocumentID` | :heavy_check_mark: | `string` | `UINT32` |
+
+- `TransactionType` indicates a new transaction type `OracleDelete`. The integer value is TBD.
```
mvadari, line 209:
```suggestion
- `TransactionType` indicates a new transaction type `OracleDelete`.
```

XLS-47d-PriceOracles/README.md
```diff
@@ -0,0 +1,403 @@
...
+###### Request JSON
+
+    {
+         "method ":  "ledger_entry ",
+         "params" : [
+             "account": "rsA2LpzuawewSBQXkiju3YQTMzW13pAAdW",
+             "oracle_document_id":  34,
+             "ledger_index ":  "validated "
+         ]
+    }
```
mvadari, line 240: If I recall the code correctly, this actually looks like this:
```suggestion
    {
         "method ":  "ledger_entry ",
         "params" : {
             "oracle: {
                 "account": "rsA2LpzuawewSBQXkiju3YQTMzW13pAAdW",
                 "oracle_document_id":  34
             },
             "ledger_index ":  "validated "
         }
    }
```

XLS-47d-PriceOracles/README.md
mvadari: It can always be changed later if we realize it's confusing, so it's not too big of a deal.

mvadari (approved):

mvadari:
Please make sure the PDF has either been updated to match the markdown or remove it.

sappenin (approved):

XLS-47d-PriceOracles/README.md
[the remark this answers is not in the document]
gregtatcam: Added a provision to delete.

XLS-47d-PriceOracles/README.md
thejohnfreeman: Yeah, I don't really care what mark is used, I just wanted something besides "no mark", which I (and I think other readers) interpret as "never required".

PR #149, opened 2023-11-20
Add CFT flag prefixes

shawnxie999:
These flag naming changes are in accordance with the actual implementation in the code.


sappenin (approved):
LGTM

PR #153, opened 2023-11-29
CFT: Remove "Free CFT" ledger reserve tier

sappenin:
Free CFTs should not be supported because they go against the rationale of having ledger reserves. Ledger reserves exist to protect the node operators from having to store free data on behalf of ledger users, without some cost. Making a "free tier" for CFTs goes against that idea, and shouldn't be modified on a per-feature basis because reserve fees aren't really related to any particular single feature. Instead, they're a ledge-wide solution to a sort of ledger-wide economic problem.

Instead, if we want to propose a fee change, we should propose that idea independent of any particular ledger feature.

Last but not least, this free tier doesn’t seem to be necessary because a user must “opt in” to hold CFTs anyway, so that user can decide at that point if they want to pay the 2XRP reserve or not (this isn't a very large amount to bear anyway).

### DIFF
--- a/XLS-0033d-compact-fungible-tokens/README.md
+++ b/XLS-0033d-compact-fungible-tokens/README.md
@@ -133,7 +133,7 @@ This value specifies the fee, in tenths of a [basis point](https://en.wikipedia.
 
 Identifies the page in the owner's directory where this item is referenced.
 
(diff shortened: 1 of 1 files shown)


shawnxie999 (approved):
Looks good. We are build on a decision later 👍

PR #169, opened 2024-01-22
Change Scale valid range to {0-20}

gregtatcam:
- Change Scale valid range to {0-20}
- Omit Scale field in SetOracle if Scale value is 0


intelliot (approved):
noted the proposed implementation sets `maxPriceScale` to 20.

https://github.com/XRPLF/rippled/pull/4789

sappenin (approved):

PR #182, opened 2024-02-22
Remove row for MPTokenIssuanceCreate flag

shawnxie999:
`lsfMPTLocked`  could not be set through a `MPTokenIssuanceCreate` transaction, as it could only be modified by a `MPTokenIssuanceSet` transaction


sappenin (approved):
LGTM, nice catch!

sappenin (approved):

PR #195, opened 2024-04-26
Update link to STAmount

sappenin:


kennyzlei (approved):

shawnxie999 (approved):

naory:
Hi, 

In the case of tickets, what should be the way to mark an MPTToken for an issued event seat as *used*?

shawnxie999:
Hi @naory, MPT can support use cases for fungible tokens, such as tickets are interchangeable with each other (eg. museum tickets that shares the same attributes). But for your particular use case where each ticket wants its own unique attributes, I think NFT might be better suited because these tickets are technically "non-fungible". For NFTs, instead of metadata, you would need to provide an IPFS link to the attributes.

naory:
Thanks for clarifying this @shawnxie999!

PR #197, opened 2024-05-06
XLS-56d: Atomic/Batch Transactions

mvadari:
Discussion thread can be found here: https://github.com/XRPLF/XRPL-Standards/discussions/162

This PR will be left open for easier commenting of specific lines/sections. The discussion will also still be available for ease of commenting for non-technical people.


XLS-0056d-batch/README.md
```diff
@@ -0,0 +1,626 @@
...
+
+The fees for the individual inner transactions are paid here instead of in the inner transaction itself, to ensure that fee escalation is calculated on the total cost of the transaction instead of just the overhead.
+
+### 2.2. `Flags`
+
+The `Flags` field represents the **batch mode** of the transaction. Exactly one must be specified in a `Batch` transaction.
+
+This spec supports four modes: 
+* `ALLORNOTHING` or `tfAllOrNothing` (with a value of `0x00000001`)
+* `ONLYONE` or `tfOnlyOne` (with a value of `0x00000002`)
```
mDuo13, line 93: I suggest renaming this mode to `FIRST` (`tfFirst`) to reflect that the order of inner transactions matters.

XLS-0056d-batch/README.md
```diff
@@ -0,0 +1,626 @@
...
+In other words, the fee is twice the base fee (a total of 20 drops when there is no fee escalation), plus the sum of the transaction fees of all the inner transactions (which incorporates factors like higher fees for multisign or `AMMCreate`).
+
+The fees for the individual inner transactions are paid here instead of in the inner transaction itself, to ensure that fee escalation is calculated on the total cost of the transaction instead of just the overhead.
+
+### 2.2. `Flags`
+
+The `Flags` field represents the **batch mode** of the transaction. Exactly one must be specified in a `Batch` transaction.
+
+This spec supports four modes: 
+* `ALLORNOTHING` or `tfAllOrNothing` (with a value of `0x00000001`)
```
mDuo13, line 92: You could optionally rename this to `ALL` (`tfAll`) which I think conveys the same idea as well or better. (The "nothing" case wouldn't really be a success in my book!)

mDuo13 (commented):
I am somewhat concerned about the greater potential for front-running with batch transactions. It seems like that is the "use case" that receives the biggest direct benefit from this feature even though we would rather discourage front-running.

Related, I think the specification should make it explicit how the batch affects the canonical order of transactions. It seems like the intent is for the canonical order to contain the outer transaction followed by the successful inner transactions in direct sequence, and that would start from wherever the outer transaction gets placed in the canonical order?

It might also be worth adding a subsection to Security about transaction fees. Specifically, make a note about how the outer transaction pays fees for _all_ the inner transactions even if using modes like `ONLYONE` where some of those transactions can fail, and how this is a precaution against exploits that could make the whole network do the work of processing candidate transactions that won't pay transaction fees.

XLS-0056d-batch/README.md
```diff
@@ -0,0 +1,626 @@
...
+
+This field operates similarly to [multisign](https://xrpl.org/docs/concepts/accounts/multi-signing/) on the XRPL. It is only needed if multiple accounts' transactions are included in the `Batch` transaction; otherwise, the normal transaction signature provides the same security guarantees.
+
+Every account that has at least one inner transaction, excluding the outer account (if applicable), must have a `BatchSigners` field.
+
+|FieldName | Required? | JSON Type | Internal Type |
+|:---------|:-----------|:---------------|:------------|
+|`Account`|✔️|`string`|`STAccount`|
+|`SigningPubKey`| |`string`|`STBlob`|
+|`Signature`| |`string`|`STBlob`|
```
tequdev, line 141: Wouldn't it be easier to understand if the field name is the same as `TxnSignature` used as CommonField?

XLS-0056d-batch/README.md
```diff
@@ -0,0 +1,626 @@
...
+        Account: "rUser2fDds782Bd6eK15RDnGMtxf7m",
+        SigningPubKey: "03C6AE25CD44323D52D28D7DE95598E6ABF953EECC9ABF767F13C21D421C034FAB",
+        Signature: "304502210083DF12FA60E2E743643889195DC42C10F62F0DE0A362330C32BBEC4D3881EECD022010579A01E052C4E587E70E5601D2F3846984DB9B16B9EBA05BAD7B51F912B899"
+      }
+    },
+  ],
+  Sequence: 4,
+  Fee: "60",
+  SigningPubKey: "03072BBE5F93D4906FC31A690A2C269F2B9A56D60DA9C2C6C0D88FB51B644C6F94",
+  TxnSignature: "30440220702ABC11419AD4940969CC32EB4D1BFDBFCA651F064F30D6E1646D74FBFC493902204E5B451B447B0F69904127F04FE71634BD825A8970B9467871DA89EEC4B021F8"
```
tequdev, line 488: If it contains Inner Transactions from multiple accounts, shouldn't the two fields `SigningPubKey` and `TxnSignature` be empty string(`""`)?

XLS-0056d-batch/README.md
```diff
@@ -0,0 +1,626 @@
...
+### A.2: What if all of the transactions fail? Will a fee still be claimed?
+
+Yes, just as they would if they were individually submitted.
+
+### A.3: Would this feature enable greater frontrunning abilities?
+
+That is definitely a concern. Ways to mitigate this are still being investigated. Some potential answers:
+* Charge for more extensive path usage
+* Have higher fees for `Batch` transactions
+* Submit the `Batch` transactions at the end of the ledger
```
mDuo13, line 588: Naively, this sounds like it would work, but in practice it might make it even easier to front-run since you could submit a mix of Batch and non-Batch transactions to more effectively "sandwich" a targeted trade.

XLS-0056d-batch/README.md
```diff
@@ -0,0 +1,626 @@
...
+
+`RawTransactions` contains the list of transactions that will be applied. There can be up to 8 transactions included. These transactions can come from one account or multiple accounts.
+
+Each inner transaction:
+* **Must** contain a `BatchTxn` field (see section 3 for details).
+* **Must not** have a sequence number. It must use a sequence number value of `0`.
+* **Must not** have a fee. It must use a fee value of `"0"`.
+* **Must not** be signed (the global transaction is already signed by all relevant parties). They must instead have an empty string (`""`) in the `SigningPubKey` and `TxnSignature` fields.
+
+A transaction will be considered a failure if it receives any result that is not `tesSUCCESS`.
```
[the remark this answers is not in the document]
mvadari: Good catch, must have gotten left behind during some of the reorgs of this spec.

XLS-0056d-batch/README.md
mvadari: That might imply to some people that it's something to do with the first transaction in the batch, rather than the first success.

XLS-0056d-batch/README.md
```diff
@@ -0,0 +1,626 @@
...
+
+## 4. Edge Cases of Transaction Processing
+
+Inner transactions don't have `Sequence`s or `TicketSequence`s, unlike a normal transaction. This causes some problems when it comes to transaction processing, due to a few edge cases.
+
+### 4.1. Ledger Object ID Generation
+
+Some objects, such as [offers](https://xrpl.org/docs/references/protocol/ledger-data/ledger-entry-types/offer/#offer-id-format) and [escrows](https://xrpl.org/docs/references/protocol/ledger-data/ledger-entry-types/escrow/#escrow-id-format), use the sequence number of the creation transaction as a part of their ledger entry ID generation, to ensure uniqueness of the IDs.
+
+To get around this, a "phantom sequence number" will be used instead. The "phantom sequence number" will be equal to `BatchTxn.TicketSequence ?? (BatchTxn.Sequence + BatchTxn.BatchIndex)`. Note that this means that any
```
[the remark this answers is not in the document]
mvadari: It's using JS syntax - `??` is a [nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing).

XLS-0056d-batch/README.md
```diff
@@ -0,0 +1,626 @@
...
+|`Signature`| |`string`|`STBlob`|
+|`Signers`| |`array`|`STArray`|
+
+#### 2.5.1. `Account`
+
+This is an account that has at least one inner transaction.
+
+#### 2.5.2. `SigningPubKey` and `Signature`
+
+These fields are included if the account is signing with a single signature (as opposed to multi-sign). They sign the `Flags` and `TxnIDs` fields.
```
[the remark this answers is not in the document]
mvadari: No, because those aren't relevant to the batching.

XLS-0056d-batch/README.md
```diff
@@ -0,0 +1,626 @@
...
+
+`TxnIDs` contains a list of the transaction hashes/IDs for all the transactions contained in `RawTransactions`. This is the only part of the inner transactions that is saved as a part of the ledger within the `Batch` transaction, since the inner transactions themselves will be their own transactions on-ledger. The hashes in `TxnIDs` **must** be in the same order as the raw transactions in `RawTransactions`.
+
+While this field seems complicated/confusing to work with, it can easily be abstracted away (e.g. as a part of autofilling) in tooling, and it's easy for `rippled` to check a hash doesn't match its corresponding transaction in `RawTransaction`.
+
+### 2.5. `BatchSigners`
+
+This field operates similarly to [multisign](https://xrpl.org/docs/concepts/accounts/multi-signing/) on the XRPL. It is only needed if multiple accounts' transactions are included in the `Batch` transaction; otherwise, the normal transaction signature provides the same security guarantees.
+
+Every account that has at least one inner transaction, excluding the outer account (if applicable), must have a `BatchSigners` field.
```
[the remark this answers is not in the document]
mvadari: Modified it, let me know if this makes sense.

XLS-0056d-batch/README.md
mvadari: It would technically have to be `TxnsSignature`, so I'd prefer leaving it as `Signature`. But it's not a strong opinion.

XLS-0056d-batch/README.md
```diff
@@ -0,0 +1,626 @@
...
+
+Some important things to note:
+* It is possible that all transactions will not be included in this list. For example, when using the `ONLYONE` mode, if the first transaction succeeds, then the rest of the transactions will not even be processed. 
+* Transactions will only be included in the ledger if their result code is `tesSUCCESS` _and_ if the outer transaction has a result code of `tesSUCCESS`. For example, the inner transaction might have a result code of `tesSUCCESS` without being included in the ledger if the `ALLORNOTHING` mode is used, but one of the transaction fails.
+
+#### 2.6.2. Inner Transactions
+
+Each inner transaction will contain the metadata for its own processing. Only the inner transactions that were actually committed to the ledger will be included. This makes it easier for legacy systems to still be able to process `Batch` transactions as if they were normal.
+
+There will also be a pointer back to the parent outer transaction (`parent_batch`), for ease of development (similar to the `nftoken_id` field).
```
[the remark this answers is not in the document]
mvadari: I was thinking added afterward, but I'm not sure if that's possible. cc @dangell7

XLS-0056d-batch/README.md
mvadari: No, the transaction isn't multisigned. The account sender still needs to sign the other fields, like `Fee`.

XLS-0056d-batch/README.md
```diff
@@ -0,0 +1,615 @@
...
+Each outer transaction will only contain the metadata for its sequence and fee processing, not for the inner transaction processing.
+
+There will also be a list of which transactions were actually processed, which is useful for the `ONLYONE` and `UNTILFAILURE` modes, since those may only process a subset of transactions, and for debugging with all modes. This section will be called `BatchExecutions`.
+
+It will contain a list of objects that have the following fields for every transaction that is processed (successfully or unsuccessfully):
+
+|FieldName | Required? | JSON Type | Internal Type |
+|:---------|:-----------|:---------------|:------------|
+|`TransactionHash`|✔️|`string`|`STUInt256`|
+|`TransactionResult`|✔️|`string`|`STBlob`|
```
dangell7, line 187: I don't think we can duplicate the field name here with another type? Maybe we use `sfBatchResult`? Also if this is an STBlob that means that the response value will be a hex. I have updated the PR.

XLS-0056d-batch/README.md
```diff
@@ -0,0 +1,615 @@
...
+
+This is the account that is submitting the outer `Batch` transaction.
+
+#### 3.1.2. `Sequence`
+
+This is the next valid sequence number of the inner transaction sender (i.e. all inner _and_ outer transactions from the same account will have the same `BatchTxn.Sequence` value). Its inclusion ensures that there are no hash collisions with other `Batch` transactions, and also allows for proper sequence number calculation (see section 4.2 for more details).
+
+In other words, a single-account `Batch` transaction will have the same sequence number in the outer transaction _and_ all the inner transactions.
+
+#### 3.1.3. `TicketSequence`
```
dangell7, line 233: Need to clarify something here. We add 1 on inner txn where outer.account == inner.account because we assume the outer txn consumed a sequence. If we use a ticket on the outer txn then we cannot add one but we wont know that based on the data from `sfBatchTxn`.

Why dont we just pre calculate the sequence? If the failure is anything but tes or tec then the txn is reverted regardless of the atomic type. If its a tes or tec it will consume the sequence. So there's not much value in calculating it in the `getSeqProxy` idk.

XLS-0056d-batch/README.md
dangell7: If we use `Signer` the the sign checks that currently exist would need to be updated/duplicated. So i'm using `TxnSignature`.

mvadari: @dangell7 just to confirm: I should update the field here to be `TxnSignature`?

XLS-0056d-batch/README.md
mvadari: What about something like `sfInnerResult`? `sfBatchResult` could be confused to mean the result of the whole `Batch`.

XLS-0056d-batch/README.md
mvadari: What do you mean by `pre-calculate the sequence`?

dangell7: Ignore that comment

XLS-0056d-batch/README.md
dangell7: `sfInnerResult` works

XLS-0056d-batch/README.md
dangell7: yes please

XLS-0056d-batch/README.md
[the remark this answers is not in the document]
dangell7: The metadata on the inner txn? We could maybe point the PreviousTxnID back to the batch? or add something similar to that?

Just to clarify you want to link the inner txn back to the outer txn?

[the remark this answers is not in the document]
mvadari: > Just to clarify you want to link the inner txn back to the outer txn?

Yes, I think that would be very useful.

XLS-0056d-batch/README.md
dangell7: We are using `BatchSigner`. Thats whats in the spec I think (JSON). I guess we could do BatchSigners/Signer and add Signers as an optional on the Signer object...  if thats what you mean.

Above I said `Signer` but meant `Signature` we are using `TxnSignature` so that the signing/verification was the same between the two (signle/multi) but I think I could use `Signature` for just the BatchSigner and leave the signers the same. 

Up to you.

XLS-0056d-batch/README.md
```diff
@@ -0,0 +1,615 @@
...
+
+## 4. Edge Cases of Transaction Processing
+
+Inner transactions don't have `Sequence`s or `TicketSequence`s, unlike a normal transaction. This causes some problems when it comes to transaction processing, due to a few edge cases.
+
+### 4.1. Ledger Object ID Generation
+
+Some objects, such as [offers](https://xrpl.org/docs/references/protocol/ledger-data/ledger-entry-types/offer/#offer-id-format) and [escrows](https://xrpl.org/docs/references/protocol/ledger-data/ledger-entry-types/escrow/#escrow-id-format), use the sequence number of the creation transaction as a part of their ledger entry ID generation, to ensure uniqueness of the IDs.
+
+To get around this, a "phantom sequence number" will be used instead. The "phantom sequence number" will be equal to `BatchTxn.TicketSequence ?? (BatchTxn.Sequence + BatchTxn.BatchIndex)`. Note that this means that any
```
shawnxie999, line 251: > Note that this means that any

The last sentence is incomplete

mvadari:
@shawnxie999 @dangell7 @mDuo13 pushed the latest changes

shawnxie999:
I think BatchExecutions in the metadata is missing in the spec

XLS-0056d-batch/README.md
```diff
@@ -0,0 +1,503 @@
...
+Each outer transaction will only contain the metadata for its sequence and fee processing, not for the inner transaction processing.
+
+There will also be a list of which transactions were actually processed, which is useful for the `ONLYONE` and `UNTILFAILURE` modes, since those may only process a subset of transactions, and for debugging with all modes. This section will be called `BatchExecutions`.
+
+It will contain a list of objects that have the following fields for every transaction that is processed (successfully or unsuccessfully):
+
+|FieldName | Required? | JSON Type | Internal Type |
+|:---------|:-----------|:---------------|:------------|
+|`TransactionHash`|✔️|`string`|`STUInt256`|
+|`InnerResult`|✔️|`string`|`STBlob`|
```
shawnxie999, line 160: doesn't it also have `sfTransactionType`?

mvadari: It currently has that in place of `TransactionHash`. I'd prefer `TransactionHash`, personally.

XLS-0056d-batch/README.md
```diff
@@ -0,0 +1,503 @@
...
+
+There will also be a pointer back to the parent outer transaction (`parent_batch`), for ease of development (similar to the `nftoken_id` field).
+
+## 3. Transaction Common Fields
+
+This standard doesn't add any new field to the [transaction common fields](https://xrpl.org/docs/references/protocol/transactions/common-fields/), but it does add another global transaction flag:
+
+| Flag Name | Value |
+|-----------|-------|
+|`tfInnerBatchTxn`|`0x40000000`|
```
ckeshava, line 178: What is the necessity of the `tfInnerBatchTxn` flag? The existence of the `parent_batch` field in the meta-data already indicates that the current txn is an inner transaction. Furthermore, the field `Fee == 0` also confirms the usage of Batch transaction.

What is the intended use of this flag?

XLS-0056d-batch/README.md
```diff
@@ -0,0 +1,503 @@
...
+It will contain a list of objects that have the following fields for every transaction that is processed (successfully or unsuccessfully):
+
+|FieldName | Required? | JSON Type | Internal Type |
+|:---------|:-----------|:---------------|:------------|
+|`TransactionHash`|✔️|`string`|`STUInt256`|
+|`InnerResult`|✔️|`string`|`STBlob`|
+
+Some important things to note:
+* It is possible that all transactions will not be included in this list. For example, when using the `ONLYONE` mode, if the first transaction succeeds, then the rest of the transactions will not even be processed. 
+* Transactions will only be included in the ledger if their result code is `tesSUCCESS` _and_ if the outer transaction has a result code of `tesSUCCESS`. For example, if the `ALLORNOTHING` mode is used and _any_ inner transaction fails, _none_ of the inner transactions will be included in the ledger.
```
ckeshava, line 164: This aspect of `Batch` transactions appears to diverge from the current understanding of the XRPL blockchain. If a non-batch transaction fails with a `tec...` error, such a transaction is included in the ledger. [Docs](https://xrpl.org/docs/references/protocol/transactions/transaction-results/tec-codes#tec-codes)

However, suppose an Outer-Batch transaction comprises of 10 inner transactions and the `tfAllOrNothing` mode. If it fails on the 9th inner transaction, then none of the transactions are included on the ledger. Am I understanding this correctly?

XLS-0056d-batch/README.md
mvadari: `Fee = 0` does not confirm the usage of `Batch`. Anyone can submit a transaction with a fee of 0.

XLS-0056d-batch/README.md
```diff
@@ -0,0 +1,503 @@
...
+`RawTransactions` contains the list of transactions that will be applied. There can be up to 8 transactions included. These transactions can come from one account or multiple accounts.
+
+Each inner transaction:
+* **Must** contain the `tfInnerBatchTxn` flag (see section 3 for details).
+* **Must not** have a fee. It must use a fee value of `"0"`.
+* **Must not** be signed (the global transaction is already signed by all relevant parties). They must instead have an empty string (`""`) in the `SigningPubKey` and `TxnSignature` fields.
+
+**This field is not included in the validated transaction, nor is it used to compute the outer transaction signature(s)**, since all transactions are included separately as a part of the ledger.
+
+### 2.4. `TxnIDs`
```
ckeshava, line 99: Can you point me to an explanation of how `TxnID` is computed in the XRPL? There are some [example usages](https://github.com/XRPLF/rippled/blob/9d58f11a60881b3edb5de80cae52d8b758ae30a8/src/libxrpl/protocol/STTx.cpp#L64) in rippled, but I didn't understand the factors contributing into `TxID`.

Larger concern: If I submit a `Batch` transaction on a sidechain, is it possible for a malicious actor to copy those signatures and re-submit the `Batch` transaction on the mainnet? Does the `TxID` hash value contain references to network_id similar to CTID (Compact Transaction Identifier) ?

ckeshava (commented):
What's the influence of `Batch` transactions on the open-ledger cost? For instance, if 10 `Batch` transactions are included in a validated ledger, this is equivalent to inserting at most 80 non-`Batch` "usual" transactions. Is the soft-limit of the open-ledger cost calculation proportionately increased by this amendment?

XLS-0056d-batch/README.md
mvadari: Also, `NetworkID` is included in the inner transactions, which is included in the hash, so it's a moot point.

XLS-0056d-batch/README.md
ckeshava: If the `Batch` transactions `tec...` codes are not included in the ledger, it leads to a break in the provenance of blockchain transactions. A few drops of XRP is irrevocably lost as transaction fees and we can never trace the reasons for such decrease in an account's balance.

Is there any design reason for excluding `tec...` transactions? What are the disadvantages?

Additionally, would the `account_tx` also not display such err-ing Batch transactions? `account_tx` uses a different database for indexing and storing the transactions and I haven't looked at the implementation.

XLS-0056d-batch/README.md
ckeshava: Such trivial transactions will be rejected by rippled due to the load-based transaction cost factor (except for `KeyReset` transactions)

mvadari: > Such trivial transactions will be rejected by rippled

And how does rippled tell the difference between one of those trivial transactions and a `Batch` transaction?

XLS-0056d-batch/README.md
mvadari: >  A few drops of XRP is irrevocably lost as transaction fees

Inner Batch transactions don't have a fee. The only fee comes from the outer transaction, which will still be validated. So no break in the provenance.

> Is there any design reason for excluding tec... transactions? What are the disadvantages?

If, as in your example, the 9th of 10 transactions with `tfAllOrNothing` fails, why is only the 9th transaction validated and not the previous 8? That's not consistent.

> Additionally, would the account_tx also not display such err-ing Batch transactions? account_tx uses a different database for indexing and storing the transactions and I haven't looked at the implementation.

It will show the whole `Batch` transaction as one, the inner transactions will not be displayed differently.

ckeshava: > Inner Batch transactions don't have a fee. The only fee comes from the outer transaction, which will still be validated. So no break in the provenance.

I missed this aspect of the spec, this clarifies my doubts 👍 

> 
> > Is there any design reason for excluding tec... transactions? What are the disadvantages?
> 
> If, as in your example, the 9th of 10 transactions with `tfAllOrNothing` fails, why is only the 9th transaction validated and not the previous 8? That's not consistent.

No, I would have preferred to specify the execution status of all the inner transactions (not just a particular failing transaction) in the ledger. This provides the best transparency for debugging purposes.

To explain my stand further, could we execute all the transactions of an `tfAllOrNothing` Batch transaction and specify the transaction-execution-result for each of them? 
> 
> > Additionally, would the account_tx also not display such err-ing Batch transactions? account_tx uses a different database for indexing and storing the transactions and I haven't looked at the implementation.
> 
> It will show the whole `Batch` transaction as one, the inner transactions will not be displayed differently.

No, you are speaking from the perspective of the account which submits the Outer-`Batch` transaction. If I query `account_tx` with one of the inner-`Batch`-transaction account's, what is the expected outcome?

I can wait for the implementation and try-it-out myself, this question is not urgent.

XLS-0056d-batch/README.md
ckeshava: ok

XLS-0056d-batch/README.md
mvadari: > No, I would have preferred to specify the execution status of all the inner transactions (not just a particular failing transaction) in the ledger. This provides the best transparency for debugging purposes.
> To explain my stand further, could we execute all the transactions of an tfAllOrNothing Batch transaction and specify the transaction-execution-result for each of them?

See the section on `BatchExecutions` (in 2.6.1).

> No, you are speaking from the perspective of the account which submits the Outer-Batch transaction. If I query account_tx with one of the inner-Batch-transaction account's, what is the expected outcome?

The inner transactions that are validated will all show up in `account_tx`. Not sure if the failed ones will show up - I don't know how the tracking of "is an account involved in this transaction" works.

> I can wait for the implementation and try-it-out myself, this question is not urgent.

https://dev.to/dangell7/batch-xls-56-is-now-available-for-testing-and-development-18bk

XLS-0056d-batch/README.md
ckeshava: Do you know which part of documentation describes the generation of TxID? I couldn't find which exact fields are included in the transaction hash generation process

XLS-0056d-batch/README.md
ckeshava: okay, that makes sense. You're trying to disambiguate an inner-`Batch` transaction from a non-`Batch` usual transaction. hmm okay

XLS-0056d-batch/README.md
mvadari: A hash is just a `SHA512` hash of the encoding of a transaction. So every field that is encoded is included.

FleetHub1988 (commented):
Merge review

XLS-0056d-batch/README.md
```diff
@@ -0,0 +1,503 @@
...
+* It is possible that all transactions will not be included in this list. For example, when using the `ONLYONE` mode, if the first transaction succeeds, then the rest of the transactions will not even be processed. 
+* Transactions will only be included in the ledger if their result code is `tesSUCCESS` _and_ if the outer transaction has a result code of `tesSUCCESS`. For example, if the `ALLORNOTHING` mode is used and _any_ inner transaction fails, _none_ of the inner transactions will be included in the ledger.
+
+#### 2.6.2. Inner Transactions
+
+Each inner transaction will contain the metadata for its own processing. Only the inner transactions that were actually committed to the ledger will be included. This makes it easier for legacy systems to still be able to process `Batch` transactions as if they were normal.
+
+There will also be a pointer back to the parent outer transaction (`parent_batch`), for ease of development (similar to the `nftoken_id` field).
+
+## 3. Transaction Common Fields
```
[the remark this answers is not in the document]
mvadari: I believe so

mvadari:
> What's the influence of `Batch` transactions on the open-ledger cost? For instance, if 10 `Batch` transactions are included in a validated ledger, this is equivalent to inserting at most 80 non-`Batch` "usual" transactions. Is the soft-limit of the open-ledger cost calculation proportionately increased by this amendment?

The open-ledger cost is tied to the total fee of all the Batch transactions, on the outer transaction.

XLS-0056d-batch/README.md
```diff
@@ -0,0 +1,535 @@
...
+This field is included if the account is signing with multi-sign (as opposed to a single signature). It operates equivalently to the [`Signers` field](https://xrpl.org/docs/references/protocol/transactions/common-fields/#signers-field) used in standard transaction multi-sign. This field holds the signatures for the `Flags` field and the hashes of the transactions in `RawTransactions`.
+
+### 2.2. Transaction Fee
+
+The fee for the outer transaction is:
+
+$$(n+2)*base\textunderscore fee + \sum_{innerTxns} Txn.Fee$$
+(where `n` is the number of signatures included in the outer transaction)
+
+In other words, the fee is twice the base fee (a total of 20 drops when there is no fee escalation), plus the sum of the transaction fees of all the inner transactions (which incorporates factors like higher fees for `AMMCreate` or `EscrowFinish`), plus
```
mDuo13, line 122: Incomplete sentence here?

mDuo13 (changes requested):
Looks good except for one incomplete sentence

XLS-0056d-batch/README.md
mvadari: Fixed, thanks for catching.

mDuo13 (approved):

XLS-0056d-batch/README.md
```diff
@@ -0,0 +1,535 @@
...
+### A.3: Would this feature enable greater frontrunning abilities?
+
+That is definitely a concern. Ways to mitigate this are still being investigated. Some potential answers:
+* Charge for more extensive path usage
+* Have higher fees for `Batch` transactions
+* Submit the `Batch` transactions at the end of the ledger
+
+### A.4: What error is returned if all the transactions fail in an `ONLYONE`/`UNTILFAILURE` transaction?
+
+A general error, `temBATCH_FAILED`/`tecBATCH_FAILED`, will be returned. A list of all the return codes encountered for the transactions that were processed will be included in the metadata, for easier debugging.
```
dangell7, line 490: This isnt quite true anymore. Inner batch tem codes are not returned and need to use the debug tool.

XLS-0056d-batch/README.md
```diff
@@ -0,0 +1,535 @@
...
+For this reason, inner transactions must include sequence numbers.
+
+### A.12: What if I want the inner transaction accounts to handle their own fees?
+
+That is not supported. This also allows fee escalation to be calculated on the total cost of the transaction, instead of just on the overhead, and ensures that even if all of the transactions fail, the transaction fees are still charged.
+
+### A.13: How will [transaction simulation](https://github.com/XRPLF/XRPL-Standards/tree/master/XLS-0069d-simulate) work with Batch?
+
+Some extra processing will be needed for that. As a result, Batch transactions likely won't be able to be simulated at first.
+
```
dangell7, line 535: Not sure if we want this here or in developer docs. 

1. Error Handling of Inner Batch Transactions is done using the debug stream. 
2. ParentBatchID is included on each of the Inner Transaction Metadata, when the result is tes or tec.

dangell7: Another point would be the cultivation when the inner transactions create and then consume an object and how the object ID is generated before the submission of the batch. Again, maybe developer docs is a better place for this.

dangell7: Same with AccountActivation, maybe just a note as it might be confusing to sign something for an account that doesnt yet exist.

XLS-0056d-batch/README.md
mvadari: Fixed

XLS-0056d-batch/README.md
mvadari: The object ID feels like a developer docs note, but I added the rest.

PR #211, opened 2024-08-02
XLS-70d: Credentials

mvadari:
Discussion thread can be found here: https://github.com/XRPLF/XRPL-Standards/discussions/202


ximinez:
Should the directory be named "simulate"?

mvadari:
> Should the directory be named "simulate"?

Oops, copy-paste error - thanks!

XLS-0070d-credentials/README.md
```diff
@@ -0,0 +1,573 @@
...
+We propose:
+* Creating a `Credential` ledger object
+* Modifying the `DepositPreauth` ledger object
+* Modifying the `DepositPreauth` transaction
+* Creating a `Credential` ledger object
+* Creating a `CredentialCreate` transaction type
+* Creating a `CredentialAccept` transaction type
+* Creating a `CredentialDelete` transaction type
+
+This feature will require an amendment, tentatively titled `featureCredentialAuth`.
```
mDuo13, line 35: Typically we don't include the `feature` prefix when naming amendments. (The prefix is used in the `rippled` codebase but typically not displayed to users nor used as part of the amendment's official name, i.e. the part that gets hashed to determine the ID or the name we use in docs.)

```suggestion
This feature will require an amendment, tentatively titled `CredentialAuth`.
```

XLS-0070d-credentials/README.md
```diff
@@ -0,0 +1,573 @@
...
+
+A [verifiable credential](https://en.wikipedia.org/wiki/Verifiable_credentials) (VC), as defined by the [W3C specification](https://www.w3.org/TR/vc-data-model-2.0/), is a secure and tamper-evident way to represent information about a subject, such as an individual, organization, or even an IoT device. These credentials are issued by a trusted entity and can be verified by third parties without directly involving the issuer at all.
+
+For a clearer understanding of the relationship between [DIDs (Decentralized Identifiers)](https://en.wikipedia.org/wiki/Decentralized_identifier) and VCs, a real-world example is considered. A DID serves as a unique identifier for an entity, similar to a fingerprint or a photo of a person (their physical characteristics). In contrast, a VC acts as a verifiable piece of information associated with that DID, much like a driver's license or passport. When a third party needs to verify an identity claim, they can examine the VC presented by the entity and ensure it aligns with the corresponding DID (such as comparing the photo on a driver's license to the physical characteristics of the person to whom it supposedly belongs).
```
mDuo13, line 41:
```suggestion
For a clearer understanding of the relationship between [DIDs (Decentralized Identifiers)](https://en.wikipedia.org/wiki/Decentralized_identifier) and VCs, consider the following analogy. A DID serves as a unique identifier for an entity, similar to a fingerprint or a photo of a person (their physical characteristics). In contrast, a VC acts as a verifiable piece of information associated with that DID, much like a driver's license or passport. When a third party needs to verify an identity claim, they can examine the VC presented by the entity and ensure it aligns with the corresponding DID (such as comparing the photo on a driver's license to the physical characteristics of the person to whom it supposedly belongs).
```

XLS-0070d-credentials/README.md
```diff
@@ -0,0 +1,573 @@
...
+
+### 1.2. Terminology
+
+* **Credential**: A representation of one or more assertions made by an issuer. For example, a passport is a representation of the assertion that a person is a citizen of a given country.
+* **Issuer**: The account that creates (issues) the credential. For example, a passport is issued by a country's government.
+* **Subject**: The account that the credential is issued to. For example, a passport is issued to a specific person.
+* **Whitelist**: A list of accounts that are allowed to do a given action, or (if used as a verb) the action of adding an account to said list.
```
mDuo13, line 48: I suggest changing "whitelist" to "allowlist" here and throughout, because "allowlist" is both more descriptive and inclusive.

XLS-0070d-credentials/README.md
```diff
@@ -0,0 +1,573 @@
...
+* **Preauthorization**: Essentially the same thing as whitelisting.
+
+*Note: These definitions do not directly match the definitions listed in the [W3C Verifiable Credentials spec](https://www.w3.org/TR/vc-data-model-2.0/#terminology). These terms are used in a slightly different way in the context of this spec.*
```
mDuo13, line 51: Howso? Aside from the presence of some terms that aren't in that list, I don't see an obvious conflict between how the terms are used here and there.

XLS-0070d-credentials/README.md
```diff
@@ -0,0 +1,573 @@
...
+*Note: These definitions do not directly match the definitions listed in the [W3C Verifiable Credentials spec](https://www.w3.org/TR/vc-data-model-2.0/#terminology). These terms are used in a slightly different way in the context of this spec.* 
+
+### 1.3. Basic Flow
+
+In this scenario, an authorizer/verifier account, Verity, is a regulated entity wants to ensure that she is only interacting with properly KYC'd accounts.
+
+* Verity sets up her account such that she only wants to interact with properly KYC'd accounts. He trusts Isabel, a credential issuer, to properly vet accounts and issue credentials attesting to that.
+* Alice, who wants to interact with Verity's account, submits her KYC documents to Isabel off-chain. Isabel creates a credential attesting to her trustworthiness on-chain.
+* Alice accepts the credential, making it a valid credential.
+* Alice can now interact with/send funds to Verity.
```
mDuo13, line 60: This section is somewhat confusingly structured and the pronouns for Verity change genders partway through. Also, typo in first sentence: missing word in "is a regulated entity [who/that] wants". I suggest first introducing the parties/roles and then describing the flow. The following is a suggestion of how you could reword it:

```suggestion
An example scenario features three parties with different roles:

* Verity is a regulated business that wants to interact only with properly KYC'd accounts, to ensure legal compliance. This makes Verity an _authorizer_ because they configure which accounts are allowed (authorized) to interact with them.
* Isabel is a credential issuer that vets accounts and issues credentials on-chain attesting that the accounts are who they say they are.
* Alice is a user who wants to interact with Verity.

The authorization flow in this scenario works as follows:

1. Verity sets up their account so that only authorized accounts can interact with it, and configures it to accept credentials issued by Isabel.
2. Alice submits her KYC documents to Isabel privately, off-chain.
3. Isabel examines Alice's documents and creates a credential on-chain attesting to Alice's trustworthiness.
4. Alice accepts the credential, making it valid.
5. Alice can now send funds to Isabel.

Importantly, the documents that Alice sends to Isabel can include personally identifiable or private information that's needed to verify Alice's identity, but this information is never published or stored on chain and Verity never needs to see it. Also, other businesses that trust Isabel can accept the same credentials so Alice does not need to repeatedly re-verify for every party she wants to interact with.
```

I also added a summary of some of the advantages of this flow for context. Hopefully those are accurate.

XLS-0070d-credentials/README.md
```diff
@@ -0,0 +1,573 @@
...
+
+## 2. On-Ledger Object: `Credential`
+
+This object will be an on-chain representation of a credential, while ensuring that any sensitive details are kept off-chain, to protect the subject's privacy.
+
+### 2.1. Fields
+
+| Field Name | Required? | JSON Type | Internal Type | Description |
+|------------|-----------|-----------|---------------|-------------|
+|`LedgerEntryType`| ✔️|`string`|`UInt16`|The ledger object's type (`Credential`).|
```
mDuo13, line 70: Nit: I prefer ✅ instead of ✔️ because it's more visible on dark backgrounds.

XLS-0070d-credentials/README.md
```diff
@@ -0,0 +1,573 @@
...
+|`Flags`| ✔️|`number`|`UInt32`|Flag values associated with this object.|
+|`Subject`| ✔️|`string`|`AccountID`|The account that the credential is for.|
+|`Issuer`| ✔️|`string`|`AccountID`|The issuer of the credential.|
+|`CredentialType`| ✔️|`string`|`Blob`|A (hex-encoded) value to identify the type of credential from the issuer.|
```
mDuo13, line 74: What is the maximum size of this field? Is an empty string allowed?

XLS-0070d-credentials/README.md
```diff
@@ -0,0 +1,573 @@
...
+#### 2.1.1. Object ID
+
+The ID of this object will be a hash that incorporates the `Subject`, `Issuer`, and `CredentialType` fields, combined with a unique space key for `Credential` objects, which will be defined during implementation.
+
+#### 2.1.2. `Flags`
+| Flag Name | Flag Value |
+|-----------|------------|
+|`lsfAccepted`|`0x00010000`|
+
+`lsfAccepted` is on if the subject of the credential has accepted the credential. If it is off, the issuer has the reserve burden; if it is on, the burden moves to the subject.
```
mDuo13, line 90: Typically in the docs we use the words "enabled" and "disabled" for flags instead of "on" and "off". Not a big deal either way though.

Instead of saying "burden", I would say "responsible" for the reserve here:

```suggestion
The `lsfAccepted` flag represents whether the subject of the credential has accepted the credential. If this flag is disabled, the issuer is responsible for this ledger entry's reserve; if the flag is enabled, the subject of the credential is responsible for the reserve instead. This flag is disabled by default, but is enabled as the result of a successful CredentialAccept transaction.
```

XLS-0070d-credentials/README.md
```diff
@@ -0,0 +1,573 @@
...
+### 1.3. Basic Flow
+
+In this scenario, an authorizer/verifier account, Verity, is a regulated entity wants to ensure that she is only interacting with properly KYC'd accounts.
+
+* Verity sets up her account such that she only wants to interact with properly KYC'd accounts. He trusts Isabel, a credential issuer, to properly vet accounts and issue credentials attesting to that.
+* Alice, who wants to interact with Verity's account, submits her KYC documents to Isabel off-chain. Isabel creates a credential attesting to her trustworthiness on-chain.
+* Alice accepts the credential, making it a valid credential.
+* Alice can now interact with/send funds to Verity.
+
+## 2. On-Ledger Object: `Credential`
```
mDuo13, line 62: I try to use the term "ledger entry" these days instead of "object" since object can be various levels.

```suggestion
## 2. `Credential` Ledger Entry
```

XLS-0070d-credentials/README.md
```diff
@@ -0,0 +1,573 @@
...
+
+This object will be an on-chain representation of a credential, while ensuring that any sensitive details are kept off-chain, to protect the subject's privacy.
```
mDuo13, line 64: Future tense isn't necessary here. Also, I suggest adding a note about the owner reserve here.

```suggestion
This ledger entry is an on-chain representation of a credential. Sensitive details are kept off-chain, to protect the subject's privacy.

This entry counts as one item for purposes of the owner reserve towards either the issuer or the subject of the credential, depending on whether the subject has accepted it.
```

XLS-0070d-credentials/README.md
```diff
@@ -0,0 +1,573 @@
...
+#### 2.1.2. `Flags`
+| Flag Name | Flag Value |
+|-----------|------------|
+|`lsfAccepted`|`0x00010000`|
+
+`lsfAccepted` is on if the subject of the credential has accepted the credential. If it is off, the issuer has the reserve burden; if it is on, the burden moves to the subject.
+
+#### 2.1.3. `CredentialType`
+
+This value is similar to the NFT `Taxon` value, where the value's meaning will be decided by the issuer. It may be the same as a claim in a VC, but could also represent a subset of such a claim.
```
mDuo13, line 94: This description feels awfully vague to me. How big can this blob be? What's an example value? It's mildly concerning to me that this field is critical to the definition of the ledger entry and yet so little of it is defined.

XLS-0070d-credentials/README.md
```diff
@@ -0,0 +1,573 @@
...
+
+| Field Name | Required? | JSON Type | Internal Type | Description |
+|------------|-----------|-----------|---------------|-------------|
+|`LedgerEntryType`| ✔️|`string`|`UInt16`|The ledger object's type (`Credential`).|
+|`Flags`| ✔️|`number`|`UInt32`|Flag values associated with this object.|
+|`Subject`| ✔️|`string`|`AccountID`|The account that the credential is for.|
+|`Issuer`| ✔️|`string`|`AccountID`|The issuer of the credential.|
+|`CredentialType`| ✔️|`string`|`Blob`|A (hex-encoded) value to identify the type of credential from the issuer.|
+|`Expiration`| |`number`|`UInt32`|Optional credential expiration.|
+|`URI`| |`string`|`Blob`|Optional additional data about the credential (such as a link to the VC document).|
```
mDuo13, line 76: I'm guessing this field is subject to the same restrictions as the one in `DID` objects, namely:

```suggestion
|`URI`| |`string`|`Blob`|Optional additional data about the credential (such as a link to the VC document). This field isn't checked for validity and is limited to a maximum length of 256 bytes. |
```

Can it be present but empty?

XLS-0070d-credentials/README.md
```diff
@@ -0,0 +1,573 @@
...
+|`OwnerNode`|✔️|`string`|`UInt64`|A hint indicating which page of the sender's owner directory links to this object, in case the directory consists of multiple pages. Note: The object does not contain a direct link to the owner directory containing it, since that value can be derived from the `Account.PreviousTxnID`.|
```
mDuo13, line 77: I think this description is copy-pasted from other places that use the same field in the docs, but it's unclear to me who the "sender" is in this context. The subject account?

On a related note, is a credential added to the owner directory of _both_ the issuer and the subject? I think it should be, even if it only counts toward the reserve of one of them.

XLS-0070d-credentials/README.md
```diff
@@ -0,0 +1,573 @@
...
+
+`lsfAccepted` is on if the subject of the credential has accepted the credential. If it is off, the issuer has the reserve burden; if it is on, the burden moves to the subject.
+
+#### 2.1.3. `CredentialType`
+
+This value is similar to the NFT `Taxon` value, where the value's meaning will be decided by the issuer. It may be the same as a claim in a VC, but could also represent a subset of such a claim.
+
+### 2.2. Account Deletion
+
+The `Credential` object is not a [deletion blocker](https://xrpl.org/docs/concepts/accounts/deleting-accounts/#requirements).
```
mDuo13, line 98: Does this mean that a `Credential` ledger entry can be automatically deleted alongside either the subject or the issuer accounts?

XLS-0070d-credentials/README.md
```diff
@@ -0,0 +1,573 @@
...
+|`OwnerNode`|✔️|`string`|`UInt64`|A hint indicating which page of the sender's owner directory links to this object, in case the directory consists of multiple pages. Note: The object does not contain a direct link to the owner directory containing it, since that value can be derived from the `Account.PreviousTxnID`.|
+|`PreviousTxnID`|✔️|`string`|`Hash256`|The identifying hash of the transaction that most recently modified this object.|
+|`PreviousTxnLgrSeqNumber`|✔️|`number`|`UInt32`|The index of the ledger that contains the transaction that most recently modified this object.|
+</details>
+
+We propose these modifications:
+
+| Field Name | Required? | JSON Type | Internal Type | Description |
+|------------|-----------|-----------|---------------|-------------|
+|`Authorize`| |`string`|`AccountID`|The account that received the preauthorization. (The sender of the preauthorized payments.)|
```
mDuo13, line 125: I would explicitly state _here_ that this field is the same but becomes optional.

XLS-0070d-credentials/README.md
```diff
@@ -0,0 +1,573 @@
...
+|`AuthorizeCredentials`| |`array`|`STArray`|The credential that received the preauthorization. (Any account with this credential can send preauthorized payments).|
```
mDuo13, line 126: This says credential (singular) but contains an array, which is confusing.

XLS-0070d-credentials/README.md
```diff
@@ -0,0 +1,573 @@
...
+
+#### 3.1.1. Object ID
+
+The ID of this object will be either a hash of the `Account` and `Authorize` fields (as it currently is), or a hash of the `Account` and the contents of `AuthorizeCredentials` fields, combined with the unique space key for `DepositAuth` objects: `0x0070`.
```
mDuo13, line 130: Is it possible that this could lead to a hash collision? I'm not concerned about the general case, but specific edge cases, like if you can craft a `AuthorizeCredentials` field and a valid account address that have the same binary representation. What would happen if you tried to authorize one and the generated ID is already used?

XLS-0070d-credentials/README.md
```diff
@@ -0,0 +1,573 @@
...
+
+This field is now optional, and a valid `DepositPreauth` object must either have the `Authorize` field or the `AuthorizeCredentials` field.
+
+#### 3.1.3. `AuthorizeCredentials`
+
+This field is an array of inner objects. The contents of these inner objects determine the credential(s) that are accepted.
+
+If more than one credential is included in the list, all of those credentials must be included (effectively ANDing them together).
+
+The list has a maximum size of 8 credentials.
```
mDuo13, line 142: Does it have a minimum size of 1? I'm guessing an empty array is not intended to be valid.

XLS-0070d-credentials/README.md
```diff
@@ -0,0 +1,573 @@
...
+|`Authorize`| |`string`|`AccountID`|The account that received the preauthorization. (The sender of the preauthorized payments.)|
+|`AuthorizeCredentials`| |`array`|`STArray`|The credential that received the preauthorization. (Any account with this credential can send preauthorized payments).|
+
+#### 3.1.1. Object ID
+
+The ID of this object will be either a hash of the `Account` and `Authorize` fields (as it currently is), or a hash of the `Account` and the contents of `AuthorizeCredentials` fields, combined with the unique space key for `DepositAuth` objects: `0x0070`.
+
+#### 3.1.2. `Authorize`
+
+This field is now optional, and a valid `DepositPreauth` object must either have the `Authorize` field or the `AuthorizeCredentials` field.
```
mDuo13, line 134: I think it would make sense to move the note about DepositPreauth requiring _either_ `Authorize` or `AuthorizeCredentials` earlier in the 3.1 section (before you get into the specific fields). Also, state explicitly that the ledger entry must have one or the other _but not both_.

mDuo13 (commented):
I've only worked my way through about half the spec so far but it mostly looks good. I've added a few comments where clarifications or rephrasing would be helpful.

XLS-0070d-credentials/README.md
```diff
@@ -0,0 +1,573 @@
...
+Note: the transaction will still fail if too many credentials are included. The exact list must be provided.
+
+### 8.3. State Changes
+
+* If the credential isn't valid, it gets blocked.
+* If the credential is expired, it is deleted.
+
+## 9. RPC: `deposit_authorized`
+
+The [`deposit_authorized` RPC method](https://xrpl.org/deposit_authorized.html) already exists on the XRPL. This proposal suggests some modifications to also support permissioned DEX domains.
```
mDuo13, line 328: "permissioned DEX domains"? Is this a leftover from a different spec?

XLS-0070d-credentials/README.md
```diff
@@ -0,0 +1,573 @@
...
+#### 5.1.2. `Signature`
+
+If the subject is submitting the account, a signature must be included from the issuer attesting to the credential. This signature will sign a simple JSON that looks something like this:
+```typescript
+{
+    Issuer: "rISABEL......",
+    Subject: "rALICE.......",
+    CredentialType: "123ABC",
+}
+```
```
mDuo13, line 224: Shouldn't the signed data also include the expiration, if one is defined? Otherwise, in the flow where the subject creates the credential using this signature, they could potentially "renew" or extend a credential by deleting it and creating a new one with a different `Expiration` value using the same old signature.

XLS-0070d-credentials/README.md
```diff
@@ -0,0 +1,573 @@
...
+
+There is no way for the XRPL to iterate through just the list of credentials an account has; it can only iterate through the entire list of account objects. This is an unbounded list of objects (it could be millions of objects). Even just the list of accepted credentials could theoretically be millions of objects long. It could cause serious performance issues if the system had to iterate through the whole list.
+
+It's much faster to have the credential ID included - it's easy to make sure that that's a valid credential, and check that it's an authorized credential.
+
+### A.7: Can a credential issuer delete their account?
+
+Yes, even if they still have issued credentials in existence.
+
+### A.8: Does a credential issue have to have an on-chain account?
```
mDuo13, line 555:
```suggestion
### A.8: Does a credential issuer have to have an on-chain account?
```

XLS-0070d-credentials/README.md
```diff
@@ -0,0 +1,573 @@
...
+    Issuer: "rISABEL......",
+    Subject: "rALICE.......",
+    CredentialType: "123ABC",
+}
+```
+
+### 5.2. Failure Conditions
+
+* Both or neither `Subject` and `Issuer` are specified (i.e. there must be only one of the two fields).
+* The account in `Subject`/`Issuer` doesn't exist.
```
mDuo13, line 229: It might actually make sense to allow you to specify unfunded accounts in the Subject/Issuer field.

This might simplify the flow in some cases where the credential issuer would otherwise have to wait for the subject to fund their XRPL account. (We previously went through something similar with Authorized Trust Lines where the token issuer sometimes wanted to authorize a trust line before the would-be-token-holder had created it.)

It might also be OK if a credential issuer doesn't actually have an on-ledger account (yet?) but they can still issue valid credentials.

XLS-0070d-credentials/README.md
```diff
@@ -0,0 +1,573 @@
...
+
+Yes, even if they still have issued credentials in existence.
+
+### A.8: Does a credential issue have to have an on-chain account?
+
+No, though they must have a valid keypair. If the second method of creating a credential is used (via a signature from the issuer), the issuer can still attest to a credential without having to have an account on-chain. This is to ensure closer compliance with the W3C Verifiable Credentials spec (see section 9).
+
+### A.9: How do I get a list of credentials that an issuer has issued?
+
+The answer to that question is still being investigated. One option would be a Clio API.
```
mDuo13, line 561: If you have the credential in both owner directories at all times, this becomes a lot easier.

It wouldn't work in cases where an issuer's account is deleted and then re-created unless you also followed the rule about deleting _all_ credentials tied to that account regardless of accepted status.

XLS-0070d-credentials/README.md
```diff
@@ -0,0 +1,573 @@
...
+
+It's much faster to have the credential ID included - it's easy to make sure that that's a valid credential, and check that it's an authorized credential.
+
+### A.7: Can a credential issuer delete their account?
+
+Yes, even if they still have issued credentials in existence.
+
+### A.8: Does a credential issue have to have an on-chain account?
+
+No, though they must have a valid keypair. If the second method of creating a credential is used (via a signature from the issuer), the issuer can still attest to a credential without having to have an account on-chain. This is to ensure closer compliance with the W3C Verifiable Credentials spec (see section 9).
```
mDuo13, line 557: I don't think this holds with the spec for `CredentialCreate` as currently defined since it states that the transaction should fail if `Issuer` specifies an account that does not exist, but you have to specify that field if you're the subject submitting a credential that was signed off-chain.

XLS-0070d-credentials/README.md
```diff
@@ -0,0 +1,573 @@
...
+
+### 5.3. State Changes
+
+If the transaction is successful:
+* The `Credential` object is created.
+* If `Issuer` == `Account`, then the `lsfAccepted` flag is enabled.
+
+## 6. Transaction: `CredentialAccept`
+
+This transaction transfers a credential from the `Issuer` to the `Account`. It is not a valid credential until it has been transferred.
```
mDuo13, line 243: I wouldn't describe it as "transferring" the credential—especially if you follow my suggestion about keeping the ledger entry in both accounts' owner directories.

```suggestion
This transaction accepts a credential issued to the `Account`. It is not a valid credential until it has been accepted.
```

mDuo13: Also: Maybe note that if you want to "un-accept" a credential you are supposed to delete it?

XLS-0070d-credentials/README.md
```diff
@@ -0,0 +1,573 @@
...
+
+This design adds support for creating, accepting, and deleting credentials. It also extends the existing [Deposit Authorization](https://xrpl.org/docs/concepts/accounts/depositauth) feature, allowing accounts to not only whitelist specific accounts, but also whitelist accounts with specific credentials.
+
+This proposal only supports blocking the interactions that Deposit Authorization supports, such as direct payments. However, future proposals could add support for credential-gating in other parts of the XRPL, such as AMMs or lending pools.
+
+We propose:
+* Creating a `Credential` ledger object
+* Modifying the `DepositPreauth` ledger object
+* Modifying the `DepositPreauth` transaction
+* Creating a `Credential` ledger object
```
mDuo13, line 30: "Creating a `Credential` ledger object" is listed twice.

Also I suggest calling it a ledger entry.

XLS-0070d-credentials/README.md
mvadari: For example, credentials can refer to many subjects, and the credentials described here aren't exactly verifiable credentials as described in the spec. Also, a subject in this case must be an on-ledger account - you can't make a credential for some off-chain entity, for example.

XLS-0070d-credentials/README.md
mvadari: I don't have a preference on whether it could be an empty string or not.

XLS-0070d-credentials/README.md
mvadari: I don't have an opinion on whether it could be an empty string or not. cc @aanchal4

XLS-0070d-credentials/README.md
mvadari: Changed to "owner".

If we want it in both owner directories, then this needs to be changed to have `SubjectNode` and `IssuerNode` fields.

XLS-0070d-credentials/README.md
mvadari: I think of "ledger entry" as referring to a specific instance, not the concept in general. But I'm not married to the term.

XLS-0070d-credentials/README.md
mvadari: Added a max length. The example value is provided below, would you rather it also be up here?

XLS-0070d-credentials/README.md
mvadari: > I think what makes the most sense is that deleting an account also deletes all Credentials where that account is responsible for the reserve, but leaves Credentials where the other party is responsible for the reserve.
> That way, even if an issuer deletes their account, valid credentials that are in use can remain in-ledger but unaccepted credentials are cleaned up.

That's basically what I meant by this section.

> I also don't think it would be totally unreasonable if deleting an account deleted all credentials linked to it regardless of accepted status.

I disagree with this, because credentials already have a mechanism for them to exist without the issuer being on-chain.

XLS-0070d-credentials/README.md
mvadari: Typo from a previous version of the spec

XLS-0070d-credentials/README.md
mvadari: I don't think that's possible, since `CredentialType` is required, but that's work checking. cc @oleks-rip

XLS-0070d-credentials/README.md
```diff
@@ -0,0 +1,573 @@
...
+
+The ID of this object will be either a hash of the `Account` and `Authorize` fields (as it currently is), or a hash of the `Account` and the contents of `AuthorizeCredentials` fields, combined with the unique space key for `DepositAuth` objects: `0x0070`.
+
+#### 3.1.2. `Authorize`
+
+This field is now optional, and a valid `DepositPreauth` object must either have the `Authorize` field or the `AuthorizeCredentials` field.
+
+#### 3.1.3. `AuthorizeCredentials`
+
+This field is an array of inner objects. The contents of these inner objects determine the credential(s) that are accepted.
```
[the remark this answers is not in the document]
mvadari: > Does the ordering of the items in the array matter?

They have to be sorted in the object. Otherwise, you'd run into issues where you provided the right group of credentials but in the wrong order, so they weren't accepted.

> Also, just to confirm, if you do have multiple DepositPreauth entries that could allow an action, it doesn't really matter which one actually did authorize it... right?

Correct. Though you'd have only two possible objects that could do so - the one with the credentials (if `CredentialIDs` is included as a field) and the one with the account.

XLS-0070d-credentials/README.md
```diff
@@ -0,0 +1,573 @@
...
+| Field Name | Required? | JSON Type | Internal Type | Description |
+|------------|-----------|-----------|---------------|-------------|
+|`Authorize`| |`string`|`AccountID`|The XRP Ledger address of the sender to preauthorize. |
+|`Unauthorize` | |`string`|`AccountID`|The XRP Ledger address of a sender whose preauthorization should be revoked.|
+
+This proposal adds two new fields:
+| Field Name | Required? | JSON Type | Internal Type | Description |
+|------------|-----------|-----------|---------------|-------------|
+|`AuthorizeCredentials`| |`array`|`STArray`|The credential(s) to preauthorize. |
+|`UnauthorizeCredentials` | |`array`|`STArray`|The credential(s) whose preauthorization should be revoked.|
```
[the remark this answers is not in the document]
mvadari: It'll be sorted when added to the ledger. So `[a,b]` and `[b,a] are functionally the same.

XLS-0070d-credentials/README.md
```diff
@@ -0,0 +1,573 @@
...
+If the transaction is successful:
+* The `DepositPreauth` object is created or deleted.
+
+## 5. Transaction: `CredentialCreate`
+
+This transaction creates a `Credential` object.
+
+There are two possible methods of doing so:
+* The issuer submits this transaction, and `CredentialAccept` is used for the subject to accept the transaction.
+* The credential data is passed off-chain from the issuer to the subject. The subject then submits the credential themselves, alongside an issuer signature, to the ledger.
```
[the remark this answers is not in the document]
mvadari: This was requested to be closer to the W3C VC spec. It doesn't expand the scope too much.

XLS-0070d-credentials/README.md
```diff
@@ -0,0 +1,573 @@
...
+|`Expiration`| |`number`|`UInt32`|Optional credential expiration.|
+|`URI`| |`string`|`Blob`|Optional additional data about the credential (such as a link to the VC document).|
+
+#### 5.1.1. `Subject` and `Issuer`
+
+Only one of these two fields must be specified. If `Subject` is specified, `Account` is the issuer. If `Issuer` is specified, `Account` is the subject.
+
+#### 5.1.2. `Signature`
+
+If the subject is submitting the account, a signature must be included from the issuer attesting to the credential. This signature will sign a simple JSON that looks something like this:
```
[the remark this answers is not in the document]
mvadari: Fixed - I think a hash prefix is an implementation detail and doesn't need to be in the spec.

XLS-0070d-credentials/README.md
mvadari: That's a good point, added.

XLS-0070d-credentials/README.md
mvadari: Oops 😨

XLS-0070d-credentials/README.md
mvadari: That's leftover from a previous version of the spec (which did not include the second method). That has been amended.

XLS-0070d-credentials/README.md
```diff
@@ -0,0 +1,573 @@
...
+
+### 8.2. Failure Conditions
+
+* Any of the `CredentialIDs` isn't an object that exists.
+* Any of the `CredentialIDs` isn't a `Credential` object.
+* Any of the `CredentialIDs` is an expired `Credential` object.
+* `CredentialIDs` is included, but the destination doesn't have Deposit Authorization set up, or the transaction doesn't have a destination (e.g. `AccountSet`).
+* The `CredentialIDs` are not authorized by the destination.
+
+Note: the transaction will still fail if too many credentials are included. The exact list must be provided.
```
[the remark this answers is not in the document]
mvadari: No, it's sorted before checking.

XLS-0070d-credentials/README.md
```diff
@@ -0,0 +1,595 @@
...
+|`TransactionType`| ✔️|`string`|`UInt16`|The transaction type (`CredentialDelete`).|
+|`Account`| ✔️|`string`|`AccountID`|The transaction submitter.|
+|`Subject`| |`string`|`AccountID`|The person that the credential is for. If omitted, `Account` is assumed to be the subject.|
+|`Issuer`| |`string`|`AccountID`|The issuer of the credential. If omitted, `Account` is assumed to be the issuer.|
+|`CredentialType`|✔️|`string`|`Blob`|A (hex-encoded) value to identify the type of credential from the issuer.|
+
+### 7.2. Failure Conditions
+
+* The credential described by the `Subject`, `Issuer`, and `CredentialType` fields doesn't exist.
+* The `Account` isn't the issuer or account, and the expiration hasn't passed.
```
ckeshava, line 308: I think you mean: `The Account isn't the Issuer or the Subject` here ^

XLS-0070d-credentials/README.md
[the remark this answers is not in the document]
ckeshava: What do you mean by "too many credentials" here? I can prove my identity through any of Passport, Driver's License or work-related-identity (ex: military card).

If there is a DepositPreAuth for such multiple credentials, then any/all of them should be acceptable.

XLS-0070d-credentials/README.md
```diff
@@ -0,0 +1,595 @@
...
+    URI: "isabel.com/credentials/kyc/alice" // This will be converted into hex
+}
+```
+
+### 12.3. `CredentialAccept`
+
+Alice accepts the credential, thereby making it valid.
+
+```typescript
+{
```
ckeshava, line 446: it would be instructive to include the updated `Flags` field in the Credential ledger object.

XLS-0070d-credentials/README.md
```diff
@@ -0,0 +1,595 @@
...
+    TransactionType: "DepositPreauth",
+    Account: "rVERITY......",
+    AuthorizeCredentials: [
+        {
+            Credential: {
+                Issuer: "rISABEL......",
+                CredentialType: "4B5943"
```
ckeshava, line 466: Suggestion: `CredentialType: "123ABC"`, Alice's credential has a different `CredentialType` than what's accepted in the DepositPreauth

XLS-0070d-credentials/README.md
```diff
@@ -0,0 +1,595 @@
...
+* The group of `CredentialIDs` is not authorized by the destination.
+
+Note: the transaction will still fail if too many credentials are included. The exact list must be provided.
+
+### 8.3. State Changes
+
+* If the credential isn't valid, the transaction fails.
+* If the credential is expired, the credential is deleted.
+
+## 9. RPC: `deposit_authorized`
```
ckeshava, line 348: Apart from the `deposit_authorized` RPC, my belief is that: no other RPC differs in the request/response format, despite the changes in the `Credential`-related content.

But certain RPCs like `account_objects` (needs to list all Credential's held by an account), `account_tx` and other `Transaction`-based RPCs have to display Credential information. Is this correct?

XLS-0070d-credentials/README.md
```diff
@@ -0,0 +1,595 @@
...
+| Field Name | Required? | JSON Type | Internal Type | Description |
+|------------|-----------|-----------|---------------|-------------|
+|`LedgerEntryType`| ✔️|`string`|`UInt16`|The ledger object's type (`Credential`).|
+|`Flags`| ✔️|`number`|`UInt32`|Flag values associated with this object.|
+|`Subject`| ✔️|`string`|`AccountID`|The account that the credential is for.|
+|`Issuer`| ✔️|`string`|`AccountID`|The issuer of the credential.|
+|`CredentialType`| ✔️|`string`|`Blob`|A (hex-encoded) value to identify the type of credential from the issuer.|
+|`Expiration`| |`number`|`UInt32`|Optional credential expiration.|
+|`URI`| |`string`|`Blob`|Optional additional data about the credential (such as a link to the VC document). This field isn't checked for validity and is limited to a maximum length of 256 bytes.|
+|`OwnerNode`|✔️|`string`|`UInt64`|A hint indicating which page of the owner's owner directory links to this object, in case the directory consists of multiple pages. Note: The object does not contain a direct link to the owner directory containing it, since that value can be derived from the `Account.PreviousTxnID`.|
```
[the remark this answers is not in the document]
mvadari: This is needed for rippled to know what page the object is. The user doesn't care. It is most often used when deleting an object.

XLS-0070d-credentials/README.md
```diff
@@ -0,0 +1,595 @@
...
+
+This field is now optional, and a valid `DepositPreauth` object must have **exactly one of** the `Authorize` field or the `AuthorizeCredentials` field.
+
+#### 3.1.3. `AuthorizeCredentials`
+
+This field is an array of inner objects. The contents of these inner objects determine the credential(s) that are accepted.
+
+If more than one credential is included in the list, all of those credentials must be included (effectively ANDing them together).
+
+The list has a minimum size of 1 and a maximum size of 8 credentials.
```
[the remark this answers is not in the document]
mvadari: No, because those objects are useless unless you're actually using the Deposit Auth flow, and there isn't really an incentive for an account to create spam just for the sake of spamming the ledger.

XLS-0070d-credentials/README.md
```diff
@@ -0,0 +1,595 @@
...
+|`Issuer`| |`string`|`AccountID`|The issuer of the credential.|
+|`CredentialType`|✔️|`string`|`Blob`|A (hex-encoded) value to identify the type of credential from the issuer.|
+|`IssuerPubKey`| |`string`|`AccountID`|The public key of the issuer.|
+|`Signature`| |`string`|`Blob`|A signature from the issuer attesting to the credential.|
+|`Expiration`| |`number`|`UInt32`|Optional credential expiration.|
+|`URI`| |`string`|`Blob`|Optional additional data about the credential (such as a link to the VC document).|
+
+#### 5.1.1. `Subject` and `Issuer`
+
+Exactly one of these two fields must be specified. If `Subject` is specified, `Account` is the issuer. If `Issuer` is specified, `Account` is the subject.
```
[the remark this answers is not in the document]
mvadari: `Account` is the account that is sending the transaction. Based on which flow is being used, a different account will be sending the transaction.

XLS-0070d-credentials/README.md
```diff
@@ -0,0 +1,595 @@
...
+* `Issuer` is included but `Signature` is not.
+* The time in `Expiration` is in the past.
+* The `URI` field is too long (limit 256 bytes).
+* The account doesn't have enough reserve for the object.
+
+### 5.3. State Changes
+
+If the transaction is successful:
+* The `Credential` object is created.
+* If `Issuer` == `Account`, then the `lsfAccepted` flag is enabled.
```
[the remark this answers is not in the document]
mvadari: Both are true. I cleaned up the wording a bit.

XLS-0070d-credentials/README.md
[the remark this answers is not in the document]
mvadari: If the `DepositPreauth` accepts two credentials, and 3 are provided, the transaction will fail.

XLS-0070d-credentials/README.md
mvadari: Yes. That doesn't need to be specified in the spec, though, since those changes are expected (most of them don't require any code changes and will automatically work).

XLS-0070d-credentials/README.md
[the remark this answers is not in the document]
sappenin: @mDuo13  -- while I agree that that this would simplify the spec and implementation, this was requested by @aanchal4 and @tapanito. In addition, our prototype already includes this feature, so removing this would be a little extra work at this point. Does that change your opinion at all @mDuo13 ?

XLS-0070d-credentials/README.md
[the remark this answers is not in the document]
ckeshava: okay

XLS-0070d-credentials/README.md
ckeshava: One of the pre-requisites for account-deletion is that the account has fewer than 1000 ledger objects. 
image


If an Issuer account has 1001 un-accepted (the subject has not sent a `CredentialAccept` transaction) Credentials, can it be deleted?

ckeshava: Opinion: From a regulatory point-of-view, non-existent (i.e. hitherto deleted) issuers might pose a problem. I do not have any legal recourse to identify my counter-party in a Payment transaction, because the Issuer has been deleted.

I know that I can choose to revoke (remove `DepositPreauth` objects) for such non-existent credentials issuers. But practically, it is difficult to monitor if Issuer accounts are deleted, in real-time.

XLS-0070d-credentials/README.md
```diff
@@ -0,0 +1,595 @@
...
+|`Expiration`| |`number`|`UInt32`|Optional credential expiration.|
+|`URI`| |`string`|`Blob`|Optional additional data about the credential (such as a link to the VC document). This field isn't checked for validity and is limited to a maximum length of 256 bytes.|
+|`OwnerNode`|✔️|`string`|`UInt64`|A hint indicating which page of the owner's owner directory links to this object, in case the directory consists of multiple pages. Note: The object does not contain a direct link to the owner directory containing it, since that value can be derived from the `Account.PreviousTxnID`.|
+|`PreviousTxnID`|✔️|`string`|`Hash256`|The identifying hash of the transaction that most recently modified this object.|
+|`PreviousTxnLgrSeqNumber`|✔️|`number`|`UInt32`|The index of the ledger that contains the transaction that most recently modified this object.|
+
+#### 2.1.1. Object ID
+
+The ID of this object will be a hash that incorporates the `Subject`, `Issuer`, and `CredentialType` fields, combined with a unique space key for `Credential` objects, which will be defined during implementation.
```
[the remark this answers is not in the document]
ckeshava: Yes, my bad

XLS-0070d-credentials/README.md
```diff
@@ -0,0 +1,600 @@
...
+
+On-chain credentialing can streamline various processes within the XRPL ecosystem. For example, financial institutions can issue credentials on the XRP Ledger, attesting to a user's identity and compliance. This eliminates the need for repeated [KYC (Know Your Customer)](https://en.wikipedia.org/wiki/Know_your_customer) checks across different platforms, fostering a smoother user experience. By enabling secure credential management, this proposal unlocks the potential for a new wave of trust-based applications within the XRPL ecosystem.
+
+## 1. Overview
+
+This design adds support for creating, accepting, and deleting credentials. It also extends the existing [Deposit Authorization](https://xrpl.org/docs/concepts/accounts/depositauth) feature, allowing accounts to not only allowlist specific accounts, but also allowlist accounts with specific credentials.
+
+This proposal only supports blocking the interactions that Deposit Authorization supports, such as direct payments. However, future proposals could add support for credential-gating in other parts of the XRPL, such as AMMs or lending pools.
```
ckeshava, line 24: `DepositAuth` functionality is similar to Authorized Trust lines for Issued Currencies. Are there any future plans to incorporate `Credential`s inside the TrustSet transaction for authorization?

mvadari: Not that I know of.

XLS-0070d-credentials/README.md
mvadari: > If an Issuer account has 1001 un-accepted (the subject has not sent a CredentialAccept transaction) Credentials, can it be deleted?

Evidently not.

> From a regulatory point-of-view, non-existent (i.e. hitherto deleted) issuers might pose a problem. I do not have any legal recourse to identify my counter-party in a Payment transaction, because the Issuer has been deleted.

The account may have been deleted, but the party still exists, and they still hold (and are responsible for) the keypair.

XLS-0070d-credentials/README.md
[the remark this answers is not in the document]
mDuo13: The prototype implementation doesn't affect my opinion on this, since the larger scope still means all of the following:

- more testing is needed
- more surface area for bugs, exploits, and surprising behavior
- more complexity for users to understand and ways for them to misunderstand or misuse the feature

For example, if you use the two-account flow to create a credential then the issuer deleting the credential is tantamount to revoking it, but this is not generally true when using the single-account flow because the subject can trivially re-create the credential. The fact that this important semantic difference emerges from the subtleties of the two flows is the type of complexity that I would like to remove by simplifying the spec.

That said, sometimes complexity is worthwhile, so if both flows have important users/use cases then it could be worth including both. On the other hand, if prospective users only want the second flow (whichever one that was), maybe the first flow can be removed. I don't have as much insight into this aspect of the design, so I leave it open to others' discretion whether both flows are needed enough to justify the complexity.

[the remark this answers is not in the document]
ckeshava: That's an interesting point. How can we prevent the Subject from re-submitting the Credential signed by the Issuer ? Are there any safe-guards against it?

XLS-0070d-credentials/README.md
```diff
@@ -0,0 +1,573 @@
...
+
+### 8.1. Fields
+
+As a reference, [here](https://xrpl.org/docs/references/protocol/transactions/common-fields/) are the fields that all transactions currently have.
+
+<!--There are too many and I didn't want to list them all, it cluttered up the spec - but maybe it can be a collapsed section?-->
+
+| Field Name | Required? | JSON Type | Internal Type |
+|------------|-----------|-----------|---------------|
+|`CredentialIDs`| ✔️|`array`|`Vector256`|Credential(s) to attach to the transaction.|
```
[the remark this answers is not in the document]
mvadari: It's the object IDs of the `Credential` objects. 

Yeah, that's exactly why - see the Appendix.

mvadari:
@sappenin this is ready to merge. The remaining open questions will be addressed in future PRs.

XLS-0070d-credentials/README.md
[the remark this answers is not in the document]
mvadari: This will be addressed in a future PR.

XLS-0070d-credentials/README.md
```diff
@@ -0,0 +1,573 @@
...
+* The account in `Account` or `Issuer` doesn't exist.
+* There is no valid credential described by the fields of the transaction.
+* The credential has already been accepted.
+* The `Account` doesn't have enough reserve for the object.
+
+### 6.3. State Changes
+
+If the transaction is successful:
+* The `lsfAccepted` flag is turned on in the credential.
+* The `Credential` object is moved from the issuer's owner directory to the subject's.
```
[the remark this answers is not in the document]
mvadari: This will be addressed in a future PR.

XLS-0070d-credentials/README.md
sappenin: Resolving this as fixed. Please re-open if you disagree.

XLS-0070d-credentials/README.md
mvadari: This change has been made.

XLS-0070d-credentials/README.md
mvadari: This will be addressed in a future PR.

sappenin (approved):

XLS-0070d-credentials/README.md
```diff
@@ -0,0 +1,602 @@
...
+#### 2.1.1. Object ID
+
+The ID of this object will be a hash that incorporates the `Subject`, `Issuer`, and `CredentialType` fields, combined with a unique space key for `Credential` objects, which will be defined during implementation.
+
+#### 2.1.2. `Flags`
+| Flag Name | Flag Value |
+|-----------|------------|
+|`lsfAccepted`|`0x00010000`|
+
+The `lsfAccepted` flag represents whether the subject of the credential has accepted the credential. If this flag is disabled, the issuer is responsible for this ledger entry's reserve; if the flag is enabled, the subject of the credential is responsible for the reserve instead. This flag is disabled by default, but is enabled as the result of a successful `CredentialAccept` transaction, or if the secondary signing flow is used to create the credential.
```
ckeshava, line 101: > , or if the secondary signing flow is used to create the credential

This sentence (at the end of the paragraph) needs to be omitted. There is only one way to create crendential objects.

mvadari: Feel free to open a PR to fix that.

XLS-0070d-credentials/README.md
```diff
@@ -0,0 +1,602 @@
...
+Note: the transaction will still fail if too many credentials are included. The exact list must be provided.
+
+### 8.3. State Changes
+
+* If the credential isn't valid, the transaction fails.
+* If the credential is expired, the credential is deleted.
+
+## 9. RPC: `deposit_authorized`
+
+The [`deposit_authorized` RPC method](https://xrpl.org/deposit_authorized.html) already exists on the XRPL. This proposal suggests some modifications to also support credential authorization.
```
tequdev, line 357: old link

https://xrpl.org/docs/references/http-websocket-apis/public-api-methods/path-and-order-book-methods/deposit_authorized

XLS-0070d-credentials/README.md
```diff
@@ -0,0 +1,602 @@
...
+
+### 9.1. Request Fields
+
+| Field Name | Required? | JSON Type | Description |
+|------------|-----------|-----------|-------------|
+|`source_account`|✔️|`string`|The sender of a possible payment.|
+|`destination_account`|✔️|`string`|The recipient of a possible payment.|
+|`ledger_hash`| |`string`|A 20-byte hex string for the ledger version to use. |
```
tequdev, line 365: 32-byte

tequdev (commented):
https://github.com/XRPLF/XRPL-Standards/pull/250

XLS-0070d-credentials/README.md
ckeshava: https://github.com/XRPLF/XRPL-Standards/pull/251

XLS-0070d-credentials/README.md
```diff
@@ -0,0 +1,602 @@
...
+We propose:
+* Creating a `Credential` ledger object
+* Modifying the `DepositPreauth` ledger object
+* Modifying the `DepositPreauth` transaction
+* Creating a `CredentialCreate` transaction type
+* Creating a `CredentialAccept` transaction type
+* Creating a `CredentialDelete` transaction type
+* Modifying other transactions that are affected by Deposit Authorization
+
+This feature will require an amendment, tentatively titled `CredentialAuth`.
```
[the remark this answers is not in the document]
mvadari: Fixed

XLS-0070d-credentials/README.md
```diff
@@ -0,0 +1,602 @@
...
+|`Flags`| ✔️|`number`|`UInt32`|Flag values associated with this object.|
+|`Subject`| ✔️|`string`|`AccountID`|The account that the credential is for.|
+|`Issuer`| ✔️|`string`|`AccountID`|The issuer of the credential.|
+|`CredentialType`| ✔️|`string`|`Blob`|A (hex-encoded) value to identify the type of credential from the issuer.|
+|`Expiration`| |`number`|`UInt32`|Optional credential expiration.|
+|`URI`| |`string`|`Blob`|Optional additional data about the credential (such as a link to the VC document). This field isn't checked for validity and is limited to a maximum length of 256 bytes.|
+|`OwnerNode`|✔️|`string`|`UInt64`|A hint indicating which page of the owner's owner directory links to this object, in case the directory consists of multiple pages. Note: The object does not contain a direct link to the owner directory containing it, since that value can be derived from the `Account.PreviousTxnID`.|
+|`PreviousTxnID`|✔️|`string`|`Hash256`|The identifying hash of the transaction that most recently modified this object.|
+|`PreviousTxnLgrSeqNumber`|✔️|`number`|`UInt32`|The index of the ledger that contains the transaction that most recently modified this object.|
```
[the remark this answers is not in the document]
mvadari: Fixed

XLS-0070d-credentials/README.md
```diff
@@ -0,0 +1,602 @@
...
+
+#### 2.1.1. Object ID
+
+The ID of this object will be a hash that incorporates the `Subject`, `Issuer`, and `CredentialType` fields, combined with a unique space key for `Credential` objects, which will be defined during implementation.
```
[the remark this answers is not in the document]
mvadari: fixed

XLS-0070d-credentials/README.md
```diff
@@ -0,0 +1,602 @@
...
+</summary>
+
+| Field Name | Required? | JSON Type | Internal Type | Description |
+|------------|-----------|-----------|---------------|-------------|
+|`Account`|✔️|`string`|`AccountID`|The account that granted the preauthorization (the destination of the preauthorized payments).|
+|`Authorize`|✔️|`string`|`AccountID`|The account that received the preauthorization (the sender of the preauthorized payments).|
+|`LedgerEntryType`|✔️|`string`|`UInt16`|The value `0x0070`, mapped to the string `"DepositPreauth"`, indicates that this is a `DepositPreauth` object.|
+|`OwnerNode`|✔️|`string`|`UInt64`|A hint indicating which page of the sender's owner directory links to this object, in case the directory consists of multiple pages. Note: The object does not contain a direct link to the owner directory containing it, since that value can be derived from the `Account.PreviousTxnID`.|
+|`PreviousTxnID`|✔️|`string`|`Hash256`|The identifying hash of the transaction that most recently modified this object.|
+|`PreviousTxnLgrSeqNumber`|✔️|`number`|`UInt32`|The index of the ledger that contains the transaction that most recently modified this object.|
```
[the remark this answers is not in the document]
mvadari: fixed

XLS-0070d-credentials/README.md
mvadari: Fixed

PR #216, opened 2024-08-16
Fix MPT implementation discrepanies

shawnxie999:
- removed upper limit of 32 issuances
- clarified MPTokenHolder field shouldn't be present 
- removed claimed fee if tfMPTCanTransfer is malformed

### DIFF
--- a/XLS-0033d-multi-purpose-tokens/README.md
+++ b/XLS-0033d-multi-purpose-tokens/README.md
@@ -509,7 +509,7 @@ Indicates the ID of the MPT involved.
 | ----------- | --------- | --------- | ------------- |
 | `MPTokenHolder` |    | `string`  | `ACCOUNTID`     | 
 
-Specifies the holders address that the issuer wants to authorize. Only used for authorization/allow-listing; must be empty if submitted by the holder.
+Specifies the holders address that the issuer wants to authorize. Only used for authorization/allow-listing; should not be present if submitted by the holder.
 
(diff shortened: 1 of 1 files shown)


XLS-0033d-multi-purpose-tokens/README.md
```diff
@@ -344,7 +344,7 @@ Specifies the flags for this transaction. In addition to the universal transacti
 
 The value specifies the fee to charged by the issuer for secondary sales of the Token, if such sales are allowed. Valid values for this field are between 0 and 50,000 inclusive, allowing transfer rates of between 0.000% and 50.000% in increments of 0.001.
 
-The field MUST NOT be present if the `tfMPTCanTransfer` flag is not set. If it is, the transaction should fail and a fee should be claimed.
+The field MUST NOT be present if the `tfMPTCanTransfer` flag is not set. If it is, the transaction should fail with `temMALFORMED`.
```
sappenin, line 347: The proposed change makes sense to me. However, it has me wondering -- what should happen if `tfMPTCanTransfer` is set, but the fee is missing? Do we assume 0 fee?

sappenin (approved):

XLS-0033d-multi-purpose-tokens/README.md
shawnxie999: Yes, the fee would be zero in this case.

gregtatcam (approved):

PR #221, opened 2024-08-30
Improve contributing guidelines

mDuo13:
Encourage discussion of invariants and error codes during the spec phase.

Some of these are based on learnings from the AMM spec and process: error codes there were changed several times during development, which led to confusion and mismatches between code, docs, and devnet as there were sometimes delays in updating any one of those with the latest changes. Also, some of the bugs in the AMM v1 implementation could have been avoided if invariants and invariant checks had been defined and implemented earlier on.

Also, `tec`-class transaction result codes are very limited, with values from 100 to 255 inclusive, and roughly a third of the existing codes have been added to the main codebase in the past 13 months, which puts us on track to run out of new `tec` codes fairly soon unless we make better efforts to conserve and reuse them. Therefore, I've added some notes which should hopefully help raise these discussions earlier in the process when changes are easier to make.

### DIFF
--- a/xls-template.md
+++ b/xls-template.md
@@ -81,9 +81,9 @@ The following is an example of how you can document new transactions, ledger ent
 
 <Any explanatory text about specific flags>
 
-For "Internal Type, most fields should use existing types defined in the XRPL binary format's type list here: https://xrpl.org/docs/references/protocol/binary-format#type-list . If a new type must be defined, add a separate section describing the rationale for the type, its binary format, and JSON representation.
+For "Internal Type", most fields should use existing types defined in the XRPL binary format's type list here: https://xrpl.org/docs/references/protocol/binary-format#type-list . If a new type must be defined, add a separate section describing the rationale for the type, its binary format, and JSON representation.
 
-For cases of transaction definitions, note any error cases that can occur. If the transaction can fail with a tec-class result code, specify which code to use. (Since tec codes are immutable data recorded in the ledger, changing them can render previous data incompatible with the current implementation. Also, tec codes are finite and limited, so it's best to reuse existing codes where appropriate.) Details of error codes can be vague or incomplete at first, but should be refined as the proposal moves through the candidate specification process.
+When defining transactions, please identify any potential error scenarios. If a transaction can fail with a `tec`-class result code, specify the appropriate code. Remember that tec codes are immutable ledger entries, so changing them can cause compatibility issues with older data. Additionally, as tec codes are limited in number, it's best to reuse existing codes whenever possible. While error code details may be initially vague or incomplete, they should be refined as the proposal progresses through the candidate specification process.
 -->
 
 ## Rationale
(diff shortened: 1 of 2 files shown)


xls-template.md
```diff
@@ -80,6 +80,10 @@ The following is an example of how you can document new object types and fields:
 >| `lsf<flag name>` | `0x0001`| <flag description> |
 
 <Any explanatory text about specific flags>
+
+For "Internal Type, most fields should use existing types defined in the XRPL binary format's type list here: https://xrpl.org/docs/references/protocol/binary-format#type-list . If a new type must be defined, add a separate section describing the rationale for the type, its binary format, and JSON representation.
```
sappenin, line 84: Not sure if this is what you meant, but it seems like a quote is missine?

```suggestion
For "Internal Type", most fields should use existing types defined in the XRPL binary format's type list here: https://xrpl.org/docs/references/protocol/binary-format#type-list . If a new type must be defined, add a separate section describing the rationale for the type, its binary format, and JSON representation.
```

xls-template.md
```diff
@@ -80,6 +80,10 @@ The following is an example of how you can document new object types and fields:
...
+
+For cases of transaction definitions, note any error cases that can occur. If the transaction can fail with a tec-class result code, specify which code to use. (Since tec codes are immutable data recorded in the ledger, changing them can render previous data incompatible with the current implementation. Also, tec codes are finite and limited, so it's best to reuse existing codes where appropriate.) Details of error codes can be vague or incomplete at first, but should be refined as the proposal moves through the candidate specification process.
```
sappenin, line 86: Feel free to ignore or adjust as you like, but this seems a bit clearer for me: 

```suggestion
When defining transactions, please identify any potential error scenarios. If a transaction can fail with a `tec`-class result code, specify the appropriate code. Remember that tec codes are immutable ledger entries, so changing them can cause compatibility issues with older data. Additionally, as tec codes are limited in number, it's best to reuse existing codes whenever possible. While error code details may be initially vague or incomplete, they should be refined as the proposal progresses through the candidate specification process.
```

xls-template.md
```diff
@@ -120,6 +124,17 @@ No backward compatibility issues found.
   TODO: Remove this comment before submitting
 -->
 
+## Invariants
+
+<!--
+  This section is optional, but recommended.
+
+  Invariants are rules for a feature's behavior that should never be broken, which define the borders of normal behavior and the assumptions involved in the design. If a situation violates an invariant, then it can be identified as unintended behavior; this helps to catch and prevent bugs. Code for the XRP Ledger includes invariant checks, which can prevent a transaction from executing if it would otherwise violate an invariant rule, preventing buggy or corrupted data from becoming part of the XRP Ledger's immutable history. The invariants defined here can be used to create invariant checks, although this spec can include invariants that are impractical to check at runtime.
```
sappenin, line 132: Feel free to ignore or adjust as you like, but this seems a bit clearer for me: 

```suggestion
Invariants are fundamental rules governing a feature's behavior that must always hold true. They define the boundaries of expected behavior and the underlying assumptions of the design. If a situation violates an invariant, it can be classified as unintended behavior, aiding in bug detection and prevention. The XRP Ledger's code incorporates invariant checks to prevent transactions from executing if they would violate an invariant rule, thereby safeguarding the ledger's immutable history from erroneous or corrupted data. While the invariants specified here can be used to create invariant checks, some may be impractical to verify at runtime.
```

sappenin (changes requested):
Thanks for these additions -- I've had "add invariants into XLS process" on my TODO list for a while, so am glad you got to it first.

sappenin (approved):
LGTM!

mvadari (approved):

intelliot (approved):

PR #223, opened 2024-08-30
Update MPT FAQ for AssetScale

sappenin:
* Update MPT FAQ for AssetScale
* Fix minor spelling errors.
* Remove TBD design in-favor of future amendment verbiage.

### DIFF
--- a/XLS-0033d-multi-purpose-tokens/README.md
+++ b/XLS-0033d-multi-purpose-tokens/README.md
@@ -836,7 +836,7 @@ Primarily because we did not want MPT meta-data (e.g., currency code, asset prec
 
 #### A.1.15. Why does `MPTokenIssuance` include an `AssetCode` instead of letting this exist solely in `MPTokenMetadata`?
 
-This specification generally prefers to have any MPT metadata reside inside of an issuer-defined `MPTokenMetadata`blob. However, `AssetScale` is an exception so that on-ledger transactors or other on-ledger components can properly perform mathematical computations on an MPT amount.
+This specification generally prefers to have any MPT metadata reside inside of an issuer-defined `MPTokenMetadata` blob. However, `AssetScale` is an exception so that on-ledger transactors or other on-ledger components can properly perform mathematical computations on an MPT amount.
 
 ### A.2. Appendix: Supplemental Information
 
--- a/XLS-0033d-multi-purpose-tokens/README.md
+++ b/XLS-0033d-multi-purpose-tokens/README.md
@@ -148,7 +148,7 @@ Specifies the sum of all token amounts that have been minted to all token holder
 
 ###### 2.1.1.2.7. `TransferFee`
 
-This value specifies the fee, in tenths of a [basis point](https://en.wikipedia.org/wiki/Basis_point), charged by the issuer for secondary sales of the token, if such sales are allowed at all. Valid values for this field are between 0 and 50,000 inclusive. A value of 1 is equivalent to 1/10 of a basis point or 0.001%, allowing transfer rates between 0% and 50%. A `TransferFee` of 50,000 corresponds to 50%. The default value for this field is 0. Any decimals in the transfer fee will be rounded down, hence the fee can be rounded down to zero if the payment is small. Issuer should make sure that their MPT's `AssetScale` is large enough.
+This value specifies the fee, in tenths of a [basis point](https://en.wikipedia.org/wiki/Basis_point), charged by the issuer for secondary sales of the token, if such sales are allowed at all. Valid values for this field are between 0 and 50,000 inclusive. A value of 1 is equivalent to 1/10 of a basis point or 0.001%, allowing transfer rates between 0% and 50%. A `TransferFee` of 50,000 corresponds to 50%. The default value for this field is 0. Any decimals in the transfer fee will be rounded down, hence the fee can be rounded down to zero if the payment is small. Issuers should make sure that their MPT's `AssetScale` is large enough.
 
 ###### 2.1.1.2.8. `PreviousTxnID`
 
@@ -792,7 +792,7 @@ After analyzing on-ledger space requirements for (a) Trustlines, (b) `MPTokenPag
 
 With all that said, this decision is still open for debate. For example, in early 2024 the Ripple team plans to perform limit testing around Trustlines and the simpler MPT design to see how increased numbers of both types of ledger objects affect ledger performance. Once that data is complete, we'll likely revisit this design choice to either validate it or change it.
 
-#### A.1.8. Why is there no `MPTRedeem` Transaction?
+#### A.1.8. Why is there no `MPTokenRedeem` Transaction?
 
 This is because holders of a MPT can use a normal payment transaction to send MPT back to the issuer, thus "redeeming" it and removing it from circulation. Note that one consequence of this design choice is that MPT issuer accounts may not also hold `MPToken` objects because, if the issuer could do such a thing, it would be ambiguous where incoming MPT payments should go (i.e., should that payment be a redemption and reduce the total amount of outstanding issuance, or should that payment go into an issuer's `MPToken` amount, and still be considered as "in circulation." For simplicity, we chose the former design, restricting MPT issuers from having `MPToken` objects at all.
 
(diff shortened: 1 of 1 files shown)


XLS-0033d-multi-purpose-tokens/README.md
```diff
@@ -834,6 +834,10 @@ Some benefits of this design approach include that is (1) reduces the footprint
 #### A.1.14. Why doesn’t an `MPTokenIssuanceID` simply hash an issuer address and currency code?
 Primarily because we did not want MPT meta-data (e.g., currency code, asset precision, common name, currency symbol, etc.) to be part of an MPTs unique identifier. This mirrors the design of other tokenization primitives on other blockchain networks that have gained massive adoption, offering strong “prior art” (e.g., ERC-20 and ERC-721 tokens). For a more detailed discussion, see [here](https://github.com/XRPLF/XRPL-Standards/discussions/128).
 
+#### A.1.15. Why does `MPTokenIssuance` include an `AssetCode` instead of letting this exist solely in `MPTokenMetadata`?
```
kennyzlei, line 837: did you mean AssetScale instead of AssetCode here?

XLS-0033d-multi-purpose-tokens/README.md
```diff
@@ -834,6 +834,10 @@ Some benefits of this design approach include that is (1) reduces the footprint
...
+
+This specification generally prefers to have any MPT metadata reside inside of an issuer-defined `MPTokenMetadata` blob. However, `AssetScale` is an exception so that on-ledger transactors or other on-ledger components can properly perform mathematical computations on an MPT amount.
+
```
shawnxie999, line 840: I don't believe the scale is involved in computations in transactors,, @gregtatcam  could you confirm?

But having an individual AssetScale field is really useful for things like the explorer, where it can display the scaled amount by dividing `amount/10^scale`. If the scale was stored in the metadata, there is really no good way to get the scale

XLS-0033d-multi-purpose-tokens/README.md
sappenin: Doh! Yes, nice catch. Fixed via 5f0fdd1632bab19a2e9b3e577674ce6c979006ac

XLS-0033d-multi-purpose-tokens/README.md
sappenin: > I don't believe the scale is involved in computations in transactors

Agreed, it's not used today as far as I can tell. However, I'm more curious about whether it will be used in the future (e.g., in the DEX or AMM?)

kennyzlei: I don't believe DEX or AMM were planning to use AssetScale either

iirc, the main rationale of having `AssetScale` to be a first class field was so web3 apps would have a standardized way to know what the scale is, where metadata was more free-form and hard to require

XLS-0033d-multi-purpose-tokens/README.md
```diff
@@ -834,6 +834,10 @@ Some benefits of this design approach include that is (1) reduces the footprint
 #### A.1.14. Why doesn’t an `MPTokenIssuanceID` simply hash an issuer address and currency code?
 Primarily because we did not want MPT meta-data (e.g., currency code, asset precision, common name, currency symbol, etc.) to be part of an MPTs unique identifier. This mirrors the design of other tokenization primitives on other blockchain networks that have gained massive adoption, offering strong “prior art” (e.g., ERC-20 and ERC-721 tokens). For a more detailed discussion, see [here](https://github.com/XRPLF/XRPL-Standards/discussions/128).
 
+#### A.1.15. Why does `MPTokenIssuance` include an `AssetScale` instead of letting this exist solely in `MPTokenMetadata`?
+
+This specification generally prefers to have any MPT metadata reside inside of an issuer-defined `MPTokenMetadata` blob. However, `AssetScale` is an exception so that on-ledger transactors or other on-ledger components can properly perform mathematical computations on an MPT amount.
```
thejohnfreeman, line 839:
```suggestion
This specification generally prefers to have MPT metadata reside inside of an issuer-defined `MPTokenMetadata` blob. However, `AssetScale` is an exception so that user interfaces can correctly format MPT values for human consumption.
```

XLS-0033d-multi-purpose-tokens/README.md
sappenin: > However, it could be argued that this field could be used on-chain in the future

Yeah, this is one key aspect I'm very interested in identifying. I don't think there are any current scenarios I can think of for "on-ledger" usage of this field, so from that perspective the wording in my PR is technically incorrect. 

However, it's possible that some future transactor (or even some future smart contract layer) might need this field, so from that perspective I think what I have in this PR should stay (but be appended -- see below).

> iirc, the main rationale of having AssetScale to be a first class field was so web3 apps would have a standardized way to know what the scale is, where metadata was more free-form and hard to require

Yes, I recall that rationale as well. In particular, I do worry about proper display and computation of MPT amounts by off-ledger applications, so how about this as an alternative for this PR:

```
This specification generally prefers to have any MPT metadata reside inside of an issuer-defined
 `MPTokenMetadata` blob. For that to work, we'll likely need to define a separate "MPT Metadata" JSON
 standard that could define a JSON payload to be embedded in `MPTokenMetadata` (sort of like [this one]
(https://eips.ethereum.org/EIPS/eip-1155#erc-1155-metadata-uri-json-schema)).

However, `AssetScale` is an exception to this guideline for three important reasons. 

1. **Future On-ledger Calculations**: While no existing on-ledger functionality (e.g., Payments, 
DEX, AMM, etc) requires `AssetScale`, it's possible that future on-ledger transactors or even future 
smart-contract layers may require this field in order to properly perform mathematical computations 
on an MPT amount.

3. **Off-ledger Amount Display**: It will be important for off-ledger applications (e.g., [XPRL Ledger Explorer]
(https://livenet.xrpl.org/)) to properly display MPT amounts in an appropriate form. For example,
 imagine a USD stablecoin issued as an MPT with a scale of `2`. In this case, a payment might 
be made for 100 units, but should probably display as `$0.01`.

4. **Off-ledger Amount Calculations**: It may be important for off-ledger applications to know a normative scale 
for an MPT, in order to make proper computations. Similar to the "display" example above, it will 
be important that an off-chain app knows how to interpret `100 units` of a given MPT in terms 
of a canonical unit value if, for example, adding that value to another similar value denominated 
in different units.
```

XLS-0033d-multi-purpose-tokens/README.md
sappenin: Great suggestion. I've incorporated this into a larger re-working of my original proposal. I'll resolve this conversation for now, but feel free to suggest changes to the current revision.

XLS-0033d-multi-purpose-tokens/README.md
sappenin: Alright, after giving it some more thought, I decided my above suggestion is too long. I spent some time this morning to (1) shorten the overall messaging; (2) define what we mean by metadata, and (3) accurately capture the several key motivations we each have around asset scale.

I'm going to resolve this conversation for now so we can focus on the latest revision. Feel free to make suggestions or add feedback on that.

XLS-0033d-multi-purpose-tokens/README.md
```diff
@@ -831,8 +831,22 @@ Last but not least, our design seeks to impose the fewest number of changes in t
...
 
-#### A.1.14. Why doesn’t an `MPTokenIssuanceID` simply hash an issuer address and currency code?
-Primarily because we did not want MPT meta-data (e.g., currency code, asset precision, common name, currency symbol, etc.) to be part of an MPTs unique identifier. This mirrors the design of other tokenization primitives on other blockchain networks that have gained massive adoption, offering strong “prior art” (e.g., ERC-20 and ERC-721 tokens). For a more detailed discussion, see [here](https://github.com/XRPLF/XRPL-Standards/discussions/128).
+#### A.1.14. Why isn’t `MPTokenIssuanceID` constructed from the hash of an issuer address and currency code?
+Primarily because we did not want MPT meta-data to be part of an MPTs unique identifier. This mirrors the design of other tokenization primitives on other blockchain networks that have gained massive adoption, offering strong “prior art” (e.g., ERC-20 and ERC-721 tokens). For a more detailed discussion, see [here](https://github.com/XRPLF/XRPL-Standards/discussions/128).
```
sappenin, line 835: Note: section `A.1.14` is unchanged aside from fixing the title.

shawnxie999 (approved):

gregtatcam (approved):
LGTM 👍

thejohnfreeman (approved):

PR #263, opened 2025-01-27
Add PermissionedDomain support for MPTs.

Tapanito:


Tapanito:
cc: @sappenin @shawnxie999

sappenin (approved):
LGTM, nice work @Tapanito !

shawnxie999:
`DomainID` would considered to be a change for another amendment, and not for XLS-33. I'm not opposing making this change (it makes things easier to find when everything is in one place). But I’m curious if there’s a formal process for "updating a previous spec based on a later spec".

shawnxie999:
`DomainID` would considered to be a change for another amendment, and not for XLS-33. I'm not opposing making this change (it makes things easier to find when everything is in one place). But I’m curious if there’s a formal process for _updating a previous spec based on a later spec_. @sappenin

shawnxie999 (approved):

mDuo13:
I think it's important to be careful/explicit about dependencies of amendments like this. What happens if MPT support is enabled on a given network and Permissioned Domains aren't?

mDuo13:
Actually, let me take this a step further and say:

**THIS SHOULD NOT BE MERGED AS-IS.**

The [MPTokensV1 amendment](https://xrpl.org/resources/known-amendments#mptokensv1) is _already_ open for voting in a released stable version of the `rippled` server. We cannot implement transaction processing changes such as this without having a _separate_ amendment for it. Otherwise, servers with the v2.3.0 version of the MPTokensV1 amendment would diverge from servers with the hypothetical future version of the amendment.

So any change to this spec should make it clear which amendment gates the addition/change to the spec.

Tapanito:
Clossing this pull-request in favor of: https://github.com/XRPLF/XRPL-Standards/pull/273

PR #274, opened 2025-03-04
XLS-64 `pseudo-account`

Tapanito:
The discussion for this specification can be found here: https://github.com/XRPLF/XRPL-Standards/discussions/191


XLS-0064d-pseudo-account/README.md
```diff
@@ -0,0 +1,94 @@
...
+
+##### 2.1.2.3. `Sequence`  
+The `Sequence` number of a` _pseudo-account_` **must** be `0`. A _pseudo-account_ cannot submit valid transactions.  
+
+##### 2.1.2.4. `RegularKey`  
+A _pseudo-account_ **must not** have a `RegularKey` set.  
+
+#### 2.1.3. Cost  
+
+A transaction that creates a `pseudo-account` must incur a higher-than-usual transaction fee to deter ledger spam. Additionally, the transaction must destroy at least the incremental owner reserve amount, currently `2 XRP`.
```
ximinez, line 85: I think section "2.1.3 Cost" should be updated to say something like:

> A pseudo-account must increase the owning / controlling account's reserve by 1 in addition to any increases required by the creation of other objects. A transaction that creates a pseudo-account that *is not owned or controlled by the creating account* must incur a higher-than-usual transaction fee to deter ledger spam, and must destroy at least the incremental owner reserve amount, currently 0.2 XRP.

I believe this spec was written based on the experiences with AMM.

From https://github.com/XRPLF/XRPL-Standards/pull/240#discussion_r2064144179
> `AMMCreate` doesn't increase anyone's owner count, partly because the object isn't owned by anyone once it's created, but that's the point - the fee is partly high because there's no one to require a reserve from.

In the case of other objects like `Vault` or `LoanBroker` where there _is_ an account that effectively owns and controls the created-pseudo-account I think it makes way more sense to increase the owner count by 2 (1 for the pseudo-account, and 1 for the vault or broker object), and charge a _normal_ transaction fee.

As a ledger object, there is nothing special or unusual about a pseudo-account if it has a single, defined owner. It's not unusually large. It doesn't function independently. It's just needed as a placeholder to allow other objects to work properly.

(Copied from https://github.com/XRPLF/XRPL-Standards/discussions/191#discussioncomment-12972650)

XLS-0064d-pseudo-account/README.md
```diff
@@ -0,0 +1,108 @@
...
+A pseudo-account must be deleted together with the associated object.
+
+###### **Invariants**
+
+The following invariants must hold true for any `AccountRoot` entry functioning as a pseudo-account:
+
+* The ledger object identified by the `<Object>ID` field must exist.
+* Exactly one `<Object>ID` field must be present on the `AccountRoot` (e.g., an account cannot be linked to both an `AMMID` and a `VaultID`).
+* The `lsfDisableMaster` and `lsfDepositAuth` flags must always be set.
+* The `Sequence` number must always be `0`.
```
ximinez:
```suggestion
* The `Sequence` number must always be `0`, and must never change.
  * AMM pseudo-accounts created under old rules will have a sequence number set to the index of the ledger they were created in. They still must never change.
```

ximinez (approved):

mvadari:
@Tapanito is this ready to go?

PR #275, opened 2025-03-13
Minor changes to Single Asset Vault specification.

Tapanito:
This commit introduces the following changes:
- Rename strFirstComeFirstServe to vaultStrategyFirstComeFirstServe
- Change asset from either string or object to only object in Vault and VaultCreate.
- Add failure condition to VaultClawback when attemptoing to clawback and asset that does not support clawback.

### DIFF
--- a/XLS-0065d-single-asset-vault/README.md
+++ b/XLS-0065d-single-asset-vault/README.md
@@ -1,6 +1,6 @@
 <pre>    
 Title:        <b>Single Asset Tokenized Vault</b>
-Revision:     <b>1</b> (2025-02-05)
+Revision:     <b>1</b> (2025-03-13)
 
 <hr>Authors:    
   <a href="mailto:vtumas@ripple.com">Vytautas Vito Tumas</a>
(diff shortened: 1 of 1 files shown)


XLS-0065d-single-asset-vault/README.md
```diff
@@ -791,3 +921,7 @@ In addion, any compliance mechanisms applied to `USDC` will also apply to the sh
 ### A-1.4 What is the difference between the `VaultOwner` and the `pseudo-account`?
 
 XRP Ledger is an account based blockchain. That means that assets (XRP, IOU and MPT must be held by an account). The Vault Object (or any other object, such as the AMM) cannot hold assets directly. Therefore, a pseudo-account is created that holds the assets on behalf of that object. The pseudo-account is a stand-alone account, that cannot receive funds, it cannot send transactions, it is there to only hold assets. So for example, when a depositor deposits assets into a vault, in reality this transaction moves the assets from the depositor account to the pseudo-account.
```
Bronek, line 923: I think the parenthesis need to be moved, like so `... (XRP, IOU and MPT) must be held by an account.` 

Would be also good to explain in this paragraph that pseudo-account also acts as an issuer of the vault shares.

XLS-0065d-single-asset-vault/README.md
```diff
@@ -770,7 +901,7 @@ We propose adding the following fields to the `ledger_entry` method:
 
 [**Return to Index**](#index)
 
-## A-1 F.A.Q.
+## A-1 F.A.Q
```
intelliot: Why remove this `.`? 😂

Tapanito: Auto formatter :laughing:

PR #290, opened 2025-06-18
XLS-26: IOU Token Metadata via xrp-ledger.toml

Mwni:
The discussion for this specification can be found here: https://github.com/XRPLF/XRPL-Standards/discussions/71

### DIFF
--- a/XLS-0026-iou-token-metadata/README.md
+++ b/XLS-0026-iou-token-metadata/README.md
@@ -43,11 +43,11 @@ Defines an issued currency on the XRPL. Currency code and issuing address must b
 |--|--|--|--|
 | `currency` 	| `string` | The currency code of the token. This can be a three-digit code, a 40-character hex code, or a custom format. | Yes
 | `issuer` 		| `string` | The public address of the issuer, encoded in the base58 format. | Yes
-| `name` 		| `string` | A display name for the token.. | No
-| `desc` 		| `string` | A description of the token and the project associated with it.. | No
-| `icon` 		| `string` | A URL pointing to an icon image representing the token. Must start with the protocol, generally _https://_ or _ipfs://_.. | No
-| `asset_class` | `string` | Top-level classification of token purpose. See the **Asset Class Definitions** table below.. | No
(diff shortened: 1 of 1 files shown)


XLS-0026-iou-token-metadata/README.md
```diff
@@ -0,0 +1,149 @@
...
+Defines an issued currency on the XRPL. Currency code and issuing address must be specified.
+| Key | Type | Description | Required
+|--|--|--|--|
+| `currency` 	| `string` | The currency code of the token. This can be a three-digit code, a 40-character hex code, or a custom format. | Yes
+| `issuer` 		| `string` | The public address of the issuer, encoded in the base58 format. | Yes
+| `name` 		| `string` | A display name for the token.. | No
+| `desc` 		| `string` | A description of the token and the project associated with it.. | No
+| `icon` 		| `string` | A URL pointing to an icon image representing the token. Must start with the protocol, generally _https://_ or _ipfs://_.. | No
+| `asset_class` | `string` | Top-level classification of token purpose. See the **Asset Class Definitions** table below.. | No
+| `asset_subclass` | `string` | Optional subcategory. See the **Asset Class Definitions** table below.. | If `rwa`
```
intelliot:
```suggestion
| `name` 		| `string` | A display name for the token. | No
| `desc` 		| `string` | A description of the token and the project associated with it. | No
| `icon` 		| `string` | A URL pointing to an icon image representing the token. Must start with the protocol, generally _https://_ or _ipfs://_. | No
| `asset_class` | `string` | Top-level classification of token purpose. See the **Asset Class Definitions** table below. | No
| `asset_subclass` | `string` | Optional subcategory. See the **Asset Class Definitions** table below. | If `rwa`
```

Mwni:
@intelliot 

> Also, why is it [[TOKENS]] rather than [[CURRENCIES]]?

The initial proposal defined both interchangeably, but the community always used TOKENS. IOUs are rarely called "currencies" by the XRP community, so I went with it. The reference implementation of the parser still [respects declared CURRENCIES](https://github.com/Mwni/xrplkit/blob/ba6959226e55461ae8d8cf6a9d4b4273b8abd4b2/xls26/xls26.js#L273).

PR #293, opened 2025-06-27
Add XLS-89d Multipurpose Token Metadata Schema

julian78780:
XLS-89d: Multipurpose Token Metadata Schema

This PR introduces XLS-89d as a Draft Specification for standardizing metadata formats used by Multipurpose Tokens (MPTs) on the XRP Ledger.

Discussion thread:
https://github.com/XRPLF/XRPL-Standards/discussions/264

Purpose:
Define a minimal, structured metadata format to support discoverability, parsing, and ecosystem-wide compatibility for MPTs across explorers, indexers, and wallets.

Key fields include:
- ticker, name, desc, asset_class, issuer_name
- Optional asset_subclass, icon, urls, and additional_info
- JSON example included

Status: Draft
Authors: Shawn Xie, Greg Tsipenyuk, Shashwat Mittal, Julian Berridi
Affiliation: Ripple

### DIFF
--- a/XLS-0089d-multi-purpose-token-metadata-schema/README.md
+++ b/XLS-0089d-multi-purpose-token-metadata-schema/README.md
@@ -1,0 +1,121 @@
+<pre>
+Title:       <b>Multi-Purpose Token Metadata Schema</b>
+Revision:    <b>Draft</b> (2025-06-27)
+
+Authors:     <a href="mailto:shawnxie@ripple.com">Shawn Xie</a>
+             <a href="mailto:gtsipenyuk@ripple.com">Greg Tsipenyuk</a>
+             <a href="mailto:smittal@ripple.com">Shashwat Mittal</a>
+             <a href="mailto:jberridi@ripple.com">Julian Berridi</a>
+
+Affiliation: <a href="https://ripple.com">Ripple</a>
+</pre>
+
+---
+
+# XLS-0089d Multi-Purpose Token Metadata Schema
+
+## 1. Abstract
+
+Multipurpose Tokens (MPTs) on the XRP Ledger allow issuers to attach arbitrary metadata to their tokens. While this flexibility is powerful, it also creates challenges for discoverability, comparability, and interoperability — both across MPTs and with existing IOUs on XRPL.
+
+This document proposes a minimally standardized metadata format for MPTs to address those challenges. The goal is not to restrict expressiveness, but to define a baseline set of fields that support reliable parsing and integration across services like block explorers, indexers, wallets, and cross-chain applications. These common fields will make it easier to surface and compare MPTs, enabling better user experiences and broader ecosystem support.
+
+This standard is optional, but MPTs that follow it will be more readily integrated into the XRPL ecosystem.
+
+---
+
+## 1.1. Approach
+
+MPTs include a 1024-byte field for arbitrary metadata. The metadata field is part of a hybrid approach of storing essential information on the ledger, while additional information can be stored off the ledger using an external URI in the metadata field. Advantages to this approach include:
+
+- Metadata can be accessed and verified on the ledger without having to fetch data from an external URI.
+- Provides increased reliability while decreasing dependency on external systems. Relying solely on external sources for metadata would introduce a dependency where services might become unavailable or unreliable, disrupting token operations and the user experience.
+- Avoids the risk of a centralized hosting service provider becoming compromised or unavailable.
+
+---
+
+## 2. Base Metadata Schema
+
+| Field             | Description                                                                                          | Example                                                                                                   | Allowed Values                                       | Type                  | Required? |
+| ----------------- | ---------------------------------------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------- | ---------------------------------------------------- | --------------------- | --------- |
+| `ticker`          | Ticker symbol used to represent the token                                                            | EXMPL                                                                                                     | Any uppercase ASCII string (max 6 chars recommended) | string                | ✔️        |
+| `name`            | Display name of the token                                                                            | Example Token                                                                                             | Any UTF-8 string                                     | string                | ✔️        |
+| `desc`            | Short description of the token                                                                       | A sample token used for demonstration                                                                     | Any UTF-8 string                                     | string                | ✔️        |
+| `icon`            | URL to the token icon                                                                                | https://example.org/token-icon.png                                                                        | HTTPS URL ending in `.png`, `.jpg`, or `.svg`        | string                | ✔️        |
+| `asset_class`     | Top-level classification of token purpose                                                            | rwa                                                                                                       | rwa, memes, wrapped, gaming, defi, other             | string                | ✔️        |
+| `asset_subclass`  | Optional subcategory, required if `asset_class = rwa`                                                | See 2.2 _asset_subclass_                                                                                  | See 2.2 _asset_subclass_                             | string                |           |
+| `issuer_name`     | The name of the issuer account                                                                       | Example Issuer                                                                                            | Any UTF-8 string                                     | string                | ✔️        |
+| `urls`            | List of related URLs (site, dashboard, social media, etc.)                                           | See 2.3 _urls_                                                                                            | See 2.3 _urls_                                       | array                 |           |
+| `additional_info` | Freeform field for key token details like interest rate, maturity date, term, or other relevant info | `{ "interest_rate": "4.75%", "maturity_date": "2030-06-30", "term": "10Y", "issuer_type": "government" }` | Any valid JSON object or UTF-8 string                | JSON object or string |           |
+
+---
+
+### 2.1 asset_class
+
+| Category  | Definition                                                                                                                                            |
+| --------- | ----------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `rwa`     | Tokens representing real-world assets (RWAs), which derive value from legally enforceable claims on physical or off-chain financial assets            |
+| `memes`   | Tokens primarily driven by community, internet culture, or speculation, without intrinsic backing or utility claims                                   |
+| `wrapped` | Tokens that represent assets from other blockchains, typically backed 1:1 and issued by bridges or custodians                                         |
+| `gaming`  | Tokens used in games or virtual worlds, often representing in-game currency, assets, or rewards                                                       |
+| `defi`    | Tokens native to or used within decentralized finance protocols, including governance tokens, DEX tokens, and lending assets                          |
+| `other`   | Tokens that do not clearly fit into the defined categories. This may include experimental, test, or those with unique use cases not covered elsewhere |
+
+---
+
+### 2.2 asset_subclass
+
+| asset_type       | Description                                                                                                                             |
+| ---------------- | --------------------------------------------------------------------------------------------------------------------------------------- |
+| `stablecoin`     | Tokens pegged to a stable value (typically fiat currencies like USD), backed by reserves such as cash, treasuries, or crypto collateral |
+| `commodity`      | Tokens representing physical commodities like gold, silver, or oil, often redeemable or legally linked to off-chain reserves            |
+| `real_estate`    | Tokens representing ownership or claims on real estate, including fractionalized property shares or REIT-like instruments               |
+| `private_credit` | Tokens representing debt obligations from private entities, such as loans, invoices, or receivables                                     |
+| `equity`         | Tokens representing ownership shares in companies, similar to traditional stock or equity instruments                                   |
+| `treasury`       | Tokens backed by or referencing government debt instruments, such as U.S. Treasury bills or bonds                                       |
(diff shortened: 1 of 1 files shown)


XLS-0089d-multi-purpose-token-metadata-schema/README.md
```diff
@@ -0,0 +1,121 @@
...
+| `other`          | Tokens that do not fit into the predefined categories above, including experimental, hybrid, or emerging real-world asset types         |
+
+---
+
+### 2.3 urls
+
+| Field   | Description                           | Example                       | Allowed Values                    | Required |
+| ------- | ------------------------------------- | ----------------------------- | --------------------------------- | -------- |
+| `url`   | The full link to the related resource | https://exampleyield.co/tbill | A valid HTTPS URL                 | ✔️       |
+| `type`  | The category of the link              | website                       | website, socialmedia, docs, other | ✔️       |
```
intelliot:
```suggestion
| `type`  | The category of the link              | website                       | website, social, document, other | ✔️       |
```
- https://github.com/XRPLF/XRPL-Standards/pull/290 uses `website`, `social`, `document` so why not keep them aligned?

XLS-0089d-multi-purpose-token-metadata-schema/README.md
```diff
@@ -0,0 +1,121 @@
...
+- Avoids the risk of a centralized hosting service provider becoming compromised or unavailable.
+
+---
+
+## 2. Base Metadata Schema
+
+| Field             | Description                                                                                          | Example                                                                                                   | Allowed Values                                       | Type                  | Required? |
+| ----------------- | ---------------------------------------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------- | ---------------------------------------------------- | --------------------- | --------- |
+| `ticker`          | Ticker symbol used to represent the token                                                            | EXMPL                                                                                                     | Any uppercase ASCII string (max 6 chars recommended) | string                | ✔️        |
```
intelliot: Are all ASCII characters permitted? Maybe restrict to A-Z, 0-9 only.

XLS-0089d-multi-purpose-token-metadata-schema/README.md
```diff
@@ -0,0 +1,121 @@
...
+| `treasury`       | Tokens backed by or referencing government debt instruments, such as U.S. Treasury bills or bonds                                       |
+| `other`          | Tokens that do not fit into the predefined categories above, including experimental, hybrid, or emerging real-world asset types         |
+
+---
+
+### 2.3 urls
+
+| Field   | Description                           | Example                       | Allowed Values                    | Required |
+| ------- | ------------------------------------- | ----------------------------- | --------------------------------- | -------- |
+| `url`   | The full link to the related resource | https://exampleyield.co/tbill | A valid HTTPS URL                 | ✔️       |
```
sappenin: We might consider adding a way for metadata to be stored off-chain, and the ledger just stores a URL link to the metadata, plus a digest of what's supposed to be there. 

We would want to think about how to make that happen though -- this might be something doable inside of a `url`, perhaps by adding an  optional field here called `digest` (whose value would be the digest of the contents of whatever content is supposed to be stored behind the url -- e.g., the contents of  `https://exampleyield.co/tbill`). This has at least two benefits:

1. Clients that are interacting with non-mutable MPTs can know if the metadata contents have changed since the MPT was created. 
2. For JSON metadata payloads that are larger than the 1024 byte limit (e.g., 2048 JSON file), then issuers could store that metadata wherever they want (e.g., at https://exampleyield.co/metadata), compute the digest, and then just store the two fields on-chain.

Of course, using the `digest+pointer` style wouldn't allow the metadata to be accessible in a transactors or in WASM, but I have always maintained that this metadata is expressly not intended for "on-ledger" consumption, but is instead meant for off-ledger consumption.

(Open to having that assumption challenged though).

XLS-0089d-multi-purpose-token-metadata-schema/README.md
sappenin: We should introduce a `type` string field that identifies what type of metadata this file represents. Something like `urn::mpt-meta::v1`. If omitted, this value could be implied to save space. 

This would have two benefits:

1. Primary: Will allow this specification to be versioned. If we ever want to add or remove fields from this spec, it will be nice for software to know what to expect.
2. Secondary: Will allow other MPT users to define their own metadata and give software a way to tell the difference.

Per my other comment below, this would also enable different profiles. For example: type=`urn::mpt-digest::v1`.

XLS-0089d-multi-purpose-token-metadata-schema/README.md
julian78780: Hey David - thanks for the suggestion — I definitely see where you're coming from, especially regarding digest+URL approaches for larger metadata payloads. That said, I think it’s worth reinforcing what we’re aiming to achieve with this proposal.

The goal here is to enable basic, structured metadata for MPTs so that they can be compared consistently — against other MPTs and IOUs. We’re purposefully keeping the scope focused: doing this one thing well, without prescribing or enforcing how more complex or off-chain metadata should be handled.

For additional details, we’re intentionally leaving the field open and flexible — issuers can use it however they see fit: off-chain pointers, detailed on-chain JSON, or otherwise. But this part is out of scope for what we're trying to standardize in v1.

Also, based on the minimum required fields we've outlined, we stay well below the 1024-byte limit. It's only when issuers choose to include extra metadata in the "additional details" that size might become a concern — and we agree that’s a valid topic, but likely better suited for a future iteration once the core format is adopted.

Appreciate the input — and we can definitely revisit this down the road as the ecosystem matures and needs become clearer.

XLS-0089d-multi-purpose-token-metadata-schema/README.md
julian78780: Good catch, agreed we should call it social instead

XLS-0089d-multi-purpose-token-metadata-schema/README.md
julian78780: Good point, will clarify

XLS-0089d-multi-purpose-token-metadata-schema/README.md
sappenin: This makes sense. I agree we can introduce something like this later. Resolving.

XLS-0089d-multi-purpose-token-metadata-schema/README.md
sappenin: Note that I think this can be added later, per the comments above by [julian78780](https://github.com/julian78780)

XLS-0089d-multi-purpose-token-metadata-schema/README.md
```diff
@@ -0,0 +1,123 @@
...
+
+# XLS-0089d Multi-Purpose Token Metadata Schema
+
+## 1. Abstract
+
+Multipurpose Tokens (MPTs) on the XRP Ledger allow issuers to attach arbitrary metadata to their tokens. While this flexibility is powerful, it also creates challenges for discoverability, comparability, and interoperability — both across MPTs and with existing IOUs on XRPL.
+
+This document proposes a minimally standardized metadata format for MPTs to address those challenges. The goal is not to restrict expressiveness, but to define a baseline set of fields that support reliable parsing and integration across services like block explorers, indexers, wallets, and cross-chain applications. These common fields will make it easier to surface and compare MPTs, enabling better user experiences and broader ecosystem support.
+
+This proposal is complementary to [PR #290](https://github.com/XRPLF/XRPL-Standards/pull/290), which defines metadata for IOUs on XRPL. While some elements differ to accommodate token-specific needs, both efforts aim to promote consistency and interoperability across the ecosystem.
```
Patel-Raj11, line 23: @julian78780
Since metadata for IOU is stored in `xrp-ledger-toml` file pointed by Domain field of AccountSet transaction, won't it make sense to have MPT's metadata in `xrp-ledger-toml` file as well? That way metadata for any token on XRPL (be it IOU or MPT) can be found at one place and it would be elegant design?

For MPT's `MPTokenMetadata` field could be a JSON with `Domain` field that points to a domain from where `xrp-ledger-toml` can be fetched similar to IOU's.

```
{
    "Domain": "ripple.com"   
}
```

Wanted your thoughts, on what prevented the design that I am describing above?

julian78780: That’s a good suggestion, and I agree it would be nice to have a unified way to surface metadata for both IOUs and MPTs. But I think the key difference is that IOUs at least include some basic context on-chain—like the currency code—so even without the .toml file, you can get a minimal sense of what the asset represents. In contrast, MPTs only store a token ID, which has no meaning without metadata.

If we were to push all MPT metadata off-chain via .toml files, we'd essentially be making the core identity of these tokens fully dependent on external infrastructure. That introduces more trust assumptions and potential fragility. With MPTs, we finally have a native, flexible way to store structured metadata directly on-chain, and I think it makes sense to lean into that—especially for critical attributes like name, symbol, and description.

That said, I can see a world where .toml files could still play a supporting role, especially for extended metadata or branding. But for the basics, I think keeping them on-chain is the right call.

mDuo13 (commented):
One question that might be worth addressing (maybe in a future extension of this spec) is any recommendations for how to convert the metadata from JSON to a binary blob that can be submitted/stored on the XRPL. There are many valid ways to do this—the most straightforward but not the most efficient would be to serialize the JSON to a UTF-8 string (whose binary representation can be attached to the transaction), but doing that will preserve the whitespace of the serialized string. You may or may not want to recommend options like serializing without extra linebreaks and indentation before converting to UTF-8.

Optimizing for minimal size before converting is probably pretty important, because the JSON Metadata example from this spec is already 742 bytes (out of a maximum 1024 bytes allowed) when serialized verbatim. Removing the whitespace reduces it to 638 bytes.

julian78780:
Good point @mDuo13 - Would it make sense to just handle this in the SDKs? We could follow your suggestion and have them serialize the JSON without line breaks or indentation before encoding to UTF-8. That way, devs can just pass in a normal object, and the SDK takes care of optimizing it under the hood. Feels like a good balance between keeping things simple for developers and doing the right thing performance wise

XLS-0089d-multi-purpose-token-metadata-schema/README.md
Patel-Raj11: Makes sense. Thanks for the context!

ckeshava:
@Patel-Raj11 
> it would create difference between the transaction that developer provided and the one that shows up on the validated ledger

Hello Raj, there are examples in today's XRPL transactions where there are differences in the representation. For example, the validated-ledgers use base-16 strings to represent arbitrary-length strings. These base-16 values are the hexadecimal representation of binary-encoded UTF-8 strings. However, using the client SDKs, developers can provide UTF-8 strings, as [shown in this example](https://github.com/XRPLF/xrpl-py/blob/c89d65b3e986caac233e49cfc960fd49ddaa3c46/tests/integration/transactions/test_set_oracle.py#L16).

ckeshava:
> Good point @mDuo13 - Would it make sense to just handle this in the SDKs? We could follow your suggestion and have them serialize the JSON without line breaks or indentation before encoding to UTF-8. That way, devs can just pass in a normal object, and the SDK takes care of optimizing it under the hood. Feels like a good balance between keeping things simple for developers and doing the right thing performance wise

@julian78780  This type of optimizations will need to be supported in both rippled + all the client libraries. I believe we need to accommodate users who send raw-JSON-transactions to a rippled node.

rippled already has a mechanism to encode arbitrary length strings (I'm referring to the `sfBlob` internal rippled type). I feel that adding another encoding-technique (outside of the peer-2-peer communication protocol) will complicate the design of the system. I'm confident that we can come up with a clever, space-saving encoding technique. However, it could incur a tech-debt-complexity cost on the dev-experience

julian78780:
Thanks everyone for the thoughtful feedback and suggestions — here’s a summary addressing the key points:

Metadata Size & Completeness
The sample metadata we’ve provided is intentionally comprehensive. It includes optional fields like description and extended structure to demonstrate a full, real-world use case. In most cases, developers will need less data to comply with the standard. It's therefore unlikely that metadata will exceed the allowed size if the standard is followed as intended

To help avoid size-related issues, we’ve added safeguards at the SDK level:
- If the metadata is too large, developers will receive a pre-submission warning before anything is sent to rippled
- The SDK will also emit console warnings if the metadata is not correctly structured (but not block the submission to rippled)
- Additionally, the description field has been made optional, giving developers more flexibility to reduce payload size if needed

Line Breaks and Indentation
We also considered the suggestion to serialize metadata without line breaks and indentation to save space. While that could result in ~20% size reduction, we believe the tradeoff in readability and maintainability isn’t justified at this point — especially since even our full sample metadata fits comfortably within the limit

That said, if we ever run into space constraints in real-world usage, we’re open to revisiting that decision in the future

Issue #294, opened 2025-07-02
Proposal: drop the "d" after the XLS number, at least for the folder names


intelliot:
It's pretty annoying that links get broken (or would get broken) when a spec moves from draft to final and the "d" gets removed from the folder name.

Let's just start off without the "d" and just indicate in the spec content itself whether it's in draft state.

sappenin:
I like this idea -- want to make a PR?

PR #298, opened 2025-07-11
more `d` removals

mvadari:
For merged/finalized specs


mvadari:
> Unfortunately, this breaks existing links. I think it's fine to keep the "d" in the spec content, but we should just always omit it from the folder name, from the outset. Then if the spec becomes finalized, its URL doesn't need to change.

I think for existing links that issue is sort of unavoidable. If we decide to go with https://github.com/XRPLF/XRPL-Standards/issues/294, we should change them all and deal with the broken links once, but not again.

PR #300, opened 2025-07-21
fix: re-establish XLS-65d as a draft

mvadari:


mvadari:
> This is ok, but to make this a non-issue, I would prefer to adopt #294 - i.e., not have "d" in the folder name, and just indicate draft in the .md file content.
> 
> If we are concerned about broken links for past specs, we can have both "d" and non-"d" folders, where the "d" folder just contains a link.

I'd like to come to a solid decision on that before making any changes about how we operate this repo. So far there's vague consensus but no official decision has been made.

PR #301, opened 2025-07-22
feat: Introduce Asset Scale for fractional asset support

Tapanito:
#### Summary

This pull request introduces a crucial new vault parameter, `AssetScale`, to enable support for fractional assets. Previously, the vault's accounting logic treated asset quantities as integers, leading to the truncation of any fractional part of a deposit and causing a loss of funds for the user.

This change refactors the core exchange algorithms to use a fixed-point arithmetic approach, where fractional assets are converted to scaled, whole-number "share" equivalents upon deposit. This ensures all internal calculations are precise and prevents value loss from rounding or truncation errors.

#### The Problem

The previous integer-based logic could not handle assets that are commonly used in fractions (e.g., ETH, WBTC). A deposit of **20.3 WETH** would be incorrectly credited as only **20 shares**, resulting in an immediate and silent loss of **0.3 WETH** for the depositor. This limitation severely restricted the types of assets the vault could support and eroded user trust.

#### The Solution

This PR implements the `AssetScale` parameter, which is defined once per vault.

1.  **Scaling on Deposit:** `AssetScale` acts as an exponent for a power-of-10 multiplier ($\sigma = 10^{\text{AssetScale}}$). During a deposit, the asset amount is multiplied by this factor to create a scaled integer representation.
    * **Example:** With an `AssetScale` of `6`, a deposit of **20.3 assets** is correctly converted to **20,300,000 shares** (`20.3 * 10^6`).

2.  **Precision in all Calculations:** All subsequent vault operations (deposits, redemptions, and withdrawals) now use these scaled share values for their calculations. This maintains precision throughout the vault's lifecycle.

3.  **Accurate Payouts:** The redeem and withdraw functions correctly interpret these scaled share values, converting them back into precise asset amounts for the user.

#### Key Changes

* **New Vault Parameter:** Added `AssetScale` to the vault's core configuration.
* **Updated Deposit Logic:** The initial deposit formula now uses the scaling factor `σ` to convert assets to shares. Subsequent calculations for deposits, redemptions, and withdrawals are updated to operate on these scaled values.
* **Rounding Consistency:** The rounding rules established in the exchange algorithms are preserved:
    * **Deposits:** Share calculations are **rounded down**.
    * **Redeems/Withdrawals:** Calculations involving asset payouts or intermediate share amounts are **rounded to the nearest number**.

### DIFF
--- a/XLS-0002-destination-information/README.md
+++ b/XLS-0002-destination-information/README.md
@@ -5,7 +5,7 @@
   author: Wietse Wind <w@xrpl-labs.com>
   discussion-from: https://github.com/XRPLF/XRPL-Standards/discussions/27
   status: Stagnant
-  category: Community
+  category: Ecosystem
   created: 2019-02-25
 </pre>
 
--- a/XLS-0003-deeplink-signed-transactions/README.md
+++ b/XLS-0003-deeplink-signed-transactions/README.md
@@ -5,7 +5,7 @@
   author: Wietse Wind <w@xrpl-labs.com>
   discussion-from: https://github.com/XRPLF/XRPL-Standards/discussions/26
   status: Stagnant
-  category: Community
+  category: Ecosystem
   created: 2019-02-25
 </pre>
 
--- a/XLS-0004-trustline-uri/README.md
+++ b/XLS-0004-trustline-uri/README.md
@@ -5,7 +5,7 @@
   author: Richard Holland (@RichardAH)
   discussion-from: https://github.com/XRPLF/XRPL-Standards/discussions/25
   status: Stagnant
-  category: Community
+  category: Ecosystem
   created: 2019-03-06
 </pre>
 
--- a/XLS-0005-tagged-addresses/README.md
+++ b/XLS-0005-tagged-addresses/README.md
@@ -5,7 +5,7 @@
   author: Nikolaos D. Bougalis <nikb@bougalis.net>
   discussion-from: https://github.com/XRPLF/XRPL-Standards/discussions/142
   status: Final
-  category: Community
+  category: Ecosystem
   created: 2019-06-04
 </pre>
 
--- a/XLS-0006-visual-account-icons/README.md
+++ b/XLS-0006-visual-account-icons/README.md
@@ -5,7 +5,7 @@
   author: Richard Holland (@RichardAH)
   discussion-from: https://github.com/XRPLF/XRPL-Standards/discussions/24
   status: Final
-  category: Community
+  category: Ecosystem
   created: 2019-09-22
 </pre>
 
--- a/XLS-0010-non-transferable-tokens/README.md
+++ b/XLS-0010-non-transferable-tokens/README.md
@@ -4,7 +4,7 @@
     author: RichardAH (@RichardAH)
     created: 2020-04-05
     status: Stagnant
-    category: Community
+    category: Ecosystem
 </pre>
 
 ## Changelog
--- a/XLS-0011-retiring-amendments/README.md
+++ b/XLS-0011-retiring-amendments/README.md
@@ -5,7 +5,7 @@
   description: This standard proposes an orderly process for retiring legacy XRP Ledger protocol behavior.
   created: 2020-05-07
   status: Final
-  category: Community
+  category: Ecosystem
 </pre>
 
 # Retiring Amendments
--- a/XLS-0012-secret-numbers/README.md
+++ b/XLS-0012-secret-numbers/README.md
@@ -5,7 +5,7 @@
   author: Wietse Wind <w@xrpl-labs.com>, Nik Bougalis (@nbougalis)
   discussion-from: https://github.com/XRPLF/XRPL-Standards/discussions/15
   status: Final
-  category: Community
+  category: Ecosystem
   created: 2020-05-13
 </pre>
 
--- a/XLS-0015-concise-tx-id/README.md
+++ b/XLS-0015-concise-tx-id/README.md
@@ -6,7 +6,7 @@
   created: 2021-03-09
   discussion-from: https://github.com/XRPLF/XRPL-Standards/discussions/34
   status: Withdrawn
-  category: Protocol
+  category: System
 </pre>
 
 ## Change Log
--- a/XLS-0016-nft-metadata/README.md
+++ b/XLS-0016-nft-metadata/README.md
@@ -4,7 +4,7 @@
     author: Hubert Getrouw (@HubertG97)
     created: 2021-03-17
     status: Stagnant
-    category: Community
+    category: Ecosystem
 </pre>
 
 In addition to @WietseWind 's [XLS-14d](https://github.com/XRPLF/XRPL-Standards/discussions/30) and @RichardAH 's proposal [XLS-15d](https://github.com/XRPLF/XRPL-Standards/discussions/34) here is a proposal to create a standard for the creation of metadata for the tokens created with a CTI in the currency code.
--- a/XLS-0017-xfl/README.md
+++ b/XLS-0017-xfl/README.md
@@ -5,7 +5,7 @@
     author: RichardAH (@RichardAH)
     created: 2021-03-19
     status: Final
-    category: Protocol
+    category: System
 </pre>
 
 # Background
--- a/XLS-0018-bootstrapping/README.md
+++ b/XLS-0018-bootstrapping/README.md
@@ -1,11 +1,11 @@
 <pre>
     xls: 18
     title: Standard For Bootstrapping XRPLD Networks
-    description: An experimental procedure to bootstrapping XRP Ledger Network 
+    description: An experimental procedure to bootstrapping XRP Ledger Network
     author: Richard Holland (@RichardAH)
     created: 2021-03-25
     status: Stagnant
-    category: Community
+    category: Ecosystem
 </pre>
 
 This procedure is experimental and additions amendments or recommendations based on experience to this standard draft are welcome.
--- a/XLS-0021-asset-code-prefixes/README.md
+++ b/XLS-0021-asset-code-prefixes/README.md
@@ -5,7 +5,7 @@
     author: Rome Reginelli (@mDuo13)
     created: 2021-07-28
     status: Stagnant
-    category: Community
+    category: Ecosystem
 </pre>
 
 # Proposal for Allocating Asset Code Prefixes
--- a/XLS-0022-api-versioning/README.md
+++ b/XLS-0022-api-versioning/README.md
@@ -4,7 +4,7 @@
   description: The API version number allows for evolving the `rippled` API while maintaining backward compatibility
   author: Elliot Lee (@intelliot), Peng Wang (@pwang200)
   status: Final
-  category: Protocol
+  category: System
   created: 2021-08-11
 </pre>
 
--- a/XLS-0024-nft-metadata/README.md
+++ b/XLS-0024-nft-metadata/README.md
@@ -5,7 +5,7 @@
     author: X-Tokenize (@x-Tokenize)
     created: 2022-02-16
     status: Final
-    category: Community
+    category: Ecosystem
 </pre>
 
 ## Introduction
--- a/XLS-0025-enhanced-secret-numbers/README.md
+++ b/XLS-0025-enhanced-secret-numbers/README.md
@@ -4,7 +4,7 @@
   description: Enhances XLS-12 secret number format by introducing an additional block for encoding ancillary information, and supporingt for longer secrets.
   author: Nik Bougalis <nikb@bougalis.net>
   status: Final
-  category: Community
+  category: Ecosystem
   created: 2021-12-10
 </pre>
 
--- a/XLS-0026-iou-token-metadata/README.md
+++ b/XLS-0026-iou-token-metadata/README.md
@@ -5,7 +5,7 @@
   author: Marc-Emanuel Otto (@Mwni)
   discussion-from: https://github.com/XRPLF/XRPL-Standards/discussions/71
   status: Draft
-  category: Community
+  category: Ecosystem
   created: 2022-04-04
 </pre>
 
--- a/XLS-0032-tx-request-uri/README.md
+++ b/XLS-0032-tx-request-uri/README.md
@@ -5,7 +5,7 @@
     description: A standard URI schema for making payments and sharing data between platforms on the XRP Ledger
     created: 2022-07-28
     status: Stagnant
-    category: Community
+    category: Ecosystem
 </pre>
 
 # Abstract
@@ -540,7 +540,7 @@ https://github.com/standardconnect/xls-32d
     <code>npm install xls-32d@latest</code>
   </p>
   <p>or</p>
-  <p> 
+  <p>
     <code>yarn add xls-32d@latest</code>
   </p>
 
--- a/XLS-0034-paychan-escrow-for-tokens/README.md
+++ b/XLS-0034-paychan-escrow-for-tokens/README.md
@@ -6,7 +6,7 @@
   discussion-from: https://github.com/XRPLF/XRPL-Standards/discussions/88
   status: Withdrawn
   withdrawal-reason: Superseded by XLS-85
-  category: Community
+  category: Ecosystem
   created: 2023-01-23
 </pre>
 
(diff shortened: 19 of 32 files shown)


XLS-0065-single-asset-vault/README.md
```diff
@@ -164,11 +165,19 @@ The `Vault` object costs one reserve fee per object created:
...
 
-##### 2.1.6.1 `MPTokenIssuance`
+##### 2.1.6.1 `AssetScale`
+
+Because vault shares are represented by MPTs, which must be whole numbers, any fractional part of a deposit gets truncated. For example, a deposit of 20.3 assets would be incorrectly credited as only 20 shares.
+
+The `AssetScale` field solves this by enabling the precise representation of fractional assets. It acts as an exponent for a power-of-10 multiplier, converting a decimal asset value into a corresponding integer share value. For instance, with an `AssetScale` of 6, a deposit of 20.3 assets is multiplied by 1,000,000 ($10^6$) and correctly stored as **20,300,000** shares.
+
+A direct consequence is that `AssetScale` also defines the minimum deposit amount, which is $10^{-\text{scale}}$. Therefore, a scale of 6 allows deposits as small as 0.000001 assets, while a scale of 0 restricts deposits to whole numbers.
+
```
Bronek, line 186: Also need to document the maximum `AssetScale` , I suggest 18 , same as `AssetScale` in MPTokenIssuance object

XLS-0065d-single-asset-vault/README.md
```diff
@@ -164,11 +165,19 @@ The `Vault` object costs one reserve fee per object created:
 
 Shares represent the portion of the Vault assets a depositor owns. Vault Owners set the currency code of the share and whether the token is transferable during the vault's creation. These two values are immutable. The share is represented by a [Multi-Purpose Token](https://github.com/XRPLF/XRPL-Standards/tree/master/XLS-0033-multi-purpose-tokens). The MPT is issued by the vault's pseudo-account.
 
-##### 2.1.6.1 `MPTokenIssuance`
+##### 2.1.6.1 `AssetScale`
+
+Because vault shares are represented by MPTs, which must be whole numbers, any fractional part of a deposit gets truncated. For example, a deposit of 20.3 assets would be incorrectly credited as only 20 shares.
```
Bronek: We do not need this part "For example, a deposit of 20.3 assets would be incorrectly credited as only 20 shares." because it does not apply after this PR is merged.

XLS-0065d-single-asset-vault/README.md
```diff
@@ -164,11 +165,19 @@ The `Vault` object costs one reserve fee per object created:
...
+
+The `AssetScale` field solves this by enabling the precise representation of fractional assets. It acts as an exponent for a power-of-10 multiplier, converting a decimal asset value into a corresponding integer share value. For instance, with an `AssetScale` of 6, a deposit of 20.3 assets is multiplied by 1,000,000 ($10^6$) and correctly stored as **20,300,000** shares.
```
Bronek: "correctly stored as" I suggest to replace this with "converted to"

XLS-0065d-single-asset-vault/README.md
```diff
@@ -164,11 +165,27 @@ The `Vault` object costs one reserve fee per object created:
...
+
+When a vault holds an **`IOU`**, the `Scale` is configurable by the Vault Owner at the time of the vault's creation. The value can range from **0** to a maximum of **18**, with a default of **6**. This flexibility allows issuers to set a level of precision appropriate for their specific token.
+
+##### 2.1.6.1.2 `XRP`
+
+When a vault holds **`XRP`**, the `Scale` is fixed at **0**. This aligns with XRP's native structure, where one share represents one "drop" (the smallest unit of XRP), and one XRP equals 1,000,000 drops. Therefore, a deposit of 10 XRP will result in the issuance of 10,000,000 shares ($10 \times 10^6$).
+
+##### 2.1.6.1.3 `MPT`
+
+When a vault holds **`XRP`**, the `Scale` is fixed at **0**. This aligns with XRP's native structure, where one share represents one "drop" (the smallest unit of XRP), and one XRP equals 1,000,000 drops. Therefore, a deposit of 10 XRP will result in the issuance of 10,000,000 shares ($10 \times 10^6$).
```
Bronek: copy & paste error - this paragraph should be about MPTs, not XRPs.

XLS-0065-single-asset-vault/README.md
```diff
@@ -164,11 +165,27 @@ The `Vault` object costs one reserve fee per object created:
 
 Shares represent the portion of the Vault assets a depositor owns. Vault Owners set the currency code of the share and whether the token is transferable during the vault's creation. These two values are immutable. The share is represented by a [Multi-Purpose Token](https://github.com/XRPLF/XRPL-Standards/tree/master/XLS-0033-multi-purpose-tokens). The MPT is issued by the vault's pseudo-account.
 
-##### 2.1.6.1 `MPTokenIssuance`
+##### 2.1.6.1 `Scale`
+
+The **`Scale`** field enables the vault to accurately represent fractional asset values using integer-only MPT shares, which prevents the loss of value from decimal truncation. It defines a scaling factor, calculated as $10^{\text{Scale}}$, that converts a decimal asset amount into a corresponding whole number of shares. For example, with an `Scale` of `6`, a deposit of **20.3** assets is multiplied by $10^6$ and credited as **20,300,000** shares.
```
Bronek: Would also add an explanation that vault shares are an MPT, and perhaps a separate paragraph (intentionally generalized, not just about shares) to explain that calculations on MPT are always carried out with the accuracy to 1 MPT where any smaller amount will be rounded in a direction which can be outside of the transaction submitter control.

Making this separate paragraph would add an extra context to the last point here which is about `Scale` for MPTs

XLS-0065d-single-asset-vault/README.md
```diff
@@ -178,8 +195,9 @@ Here’s the table with the headings "Field," "Description," and "Value":
 | `MaximumAmount`   | No limit to the number of shares that can be issued.   | `0xFFFFFFFFFFFFFFFF` |
 | `TransferFee`     | The fee paid to transfer the shares.                   | 0                    |
 | `MPTokenMetadata` | Arbitrary metadata about the share MPT, in hex format. | -                    |
+| `Scale`      | Represents the number of orders of magnitude between a standard unit and an MPT unit | `Vault.Scale` |
```
Bronek: This is incorrect - the new `Scale` field is only added to `Vault` object, not to `MPTokenIssuance`. Other users of `MPTokenIssuance` would have no use for it, and also there is no need to store it in both locations.

The `MPTokenIssuance` already has `AssetScale` which will be populated from `Scale` upon the creation of the vault, but is only ever used as a metadata (i.e. not directly used in XRPL) and is only meant for display purposes by the ledger users.

XLS-0065d-single-asset-vault/README.md
```diff
@@ -344,6 +394,10 @@ The transaction creates an `AccountRoot` object for the `_pseudo-account_`. Ther
 
 ##### 3.1.1.4 Failure Conditions
 
+- The `Asset` is `XRP`:
+
+  The `Scale` parameter is provided, and not equal to **6**. (`Scale` must be **6** for XRP.)
```
Bronek: This is incorrect. For both `XRP` and `MPT`, `Scale` is fixed at 0 ; providing this parameter in `VaultCreate` (even if it is set to zero) is an error.

XLS-0065d-single-asset-vault/README.md
```diff
@@ -312,16 +361,17 @@ The `Vault` object is managed with `VaultCreate`, `VaultSet` and `VaultDelete` t
...
+| `AssetsMaximum`     |                    |      `number`      |   `NUMBER`    |            0             | The maximum asset amount that can be held in a vault.                           |
+| `MPTokenMetadata`  |                    |      `string`      |    `BLOB`     |                          | Arbitrary metadata about the share `MPT`, in hex format, limited to 1024 bytes. |
+| `WithdrawalPolicy` |                    |      `number`      |    `UINT8`    | `strFirstComeFirstServe` | Indicates the withdrawal strategy used by the Vault.                            |
+| `DomainID`         |                    |      `string`      |   `HASH256`   |                          | The `PermissionedDomain` object ID associated with the shares of this Vault.    |
+| `Scale`       |                    |      `number`      |    `UINT8`    |            6             | The `Scale` specifies the power of 10 ($10^{\text{scale}}$) to multiply an asset's value by when converting it into an integer-based number of shares. |
```
Bronek: I would also add a comment that `VaultCreate` transaction will automatically create an `MPToken` object for vault shares for the vault creator.

XLS-0065d-single-asset-vault/README.md
Tapanito: This is a naming issue on my end, it should've read `AssetScale` , which is populated from the `Vault.Scale`

XLS-0065d-single-asset-vault/README.md
Tapanito: Thanks! Updated and added.

XLS-0065d-single-asset-vault/README.md
Tapanito: Done, added this to state changes.

XLS-0065-single-asset-vault/README.md
Tapanito: Good call, added!

XLS-0065-single-asset-vault/README.md
```diff
@@ -169,17 +169,19 @@ Shares represent the portion of the Vault assets a depositor owns. Vault Owners
...
+When a vault holds **`XRP`**, the `Scale` is fixed at **0**. This aligns with XRP's native structure, where one share represents one "drop" (the smallest unit of XRP), and one XRP equals 1,000,000 drops. Therefore, a deposit of 10 XRP to an empty Vault will result in the issuance of 10,000,000 shares ($10 \times 10^6$).
 
 ##### 2.1.6.1.3 `MPT`
 
-When a vault holds **`XRP`**, the `Scale` is fixed at **0**. This aligns with XRP's native structure, where one share represents one "drop" (the smallest unit of XRP), and one XRP equals 1,000,000 drops. Therefore, a deposit of 10 XRP will result in the issuance of 10,000,000 shares ($10 \times 10^6$).
+When a vault holds `MPT`, its `Scale` is fixed at **0**. This creates a 1-to-1 relationship between deposited MPT units and the shares issued (for example, depositing 10 MPTs to an empty Vault issues 10 shares). The value of a single MPT is determined at the issuer's discretion. If an MPT is set to represent a large value, the vault owner must be cautious. Because only whole MPT units can be deposited, any value that is not a multiple of a single MPT's value may be lost due to rounding during a transaction.
```
Bronek: "the vault owner and depositors should be cautious. Since only whole MPT units are used in calculations, any value that is ..."

XLS-0065d-single-asset-vault/README.md
```diff
@@ -189,22 +191,22 @@ The `MPTokenIssuance` object represents the share on the ledger. It is created a
...
-| `Scale`      | Represents the number of orders of magnitude between a standard unit and an MPT unit | `Vault.Scale` |
+| **Field**         | **Description**                                                                                 | **Value**            |
+| ----------------- | ----------------------------------------------------------------------------------------------- | -------------------- |
+| `Issuer`          | The AccountID of the Vault's _pseudo-account_.                                                  | _pseudo-account_ ID  |
+| `MaximumAmount`   | No limit to the number of shares that can be issued.                                            | `0xFFFFFFFFFFFFFFFF` |
+| `TransferFee`     | The fee paid to transfer the shares.                                                            | 0                    |
+| `MPTokenMetadata` | Arbitrary metadata about the share MPT, in hex format.                                          | -                    |
+| `AssetScale`      | Metadata representing the number of orders of magnitude between a standard unit and an MPT unit | `Vault.Scale`        |
```
Bronek: would also comment "only if the vault asset is an IOU", somewhere (probably not in the table ?)

XLS-0065-single-asset-vault/README.md
```diff
@@ -344,12 +397,17 @@ The transaction creates an `AccountRoot` object for the `_pseudo-account_`. Ther
...
+  - The `Scale` parameter is provided.
+
 - The `Asset` is `MPT`:
+  - The `Scale` parameter is provided.
   - The `lsfMPTCanTransfer` is not set in the `MPTokenIssuance` object. (the asset is not transferable).
   - The `lsfMPTLocked` flag is set in the `MPTokenIssuance` object. (the asset is locked).
 
 - The `Asset` is an `IOU`:
   - The `lsfGlobalFreeze` flag is set on the issuing account (the asset is frozen).
+    The `Scale` parameter is provided, and is less than **0** or greater than **18**.
```
Bronek: I think this line is missing ` -` at the start.

XLS-0065-single-asset-vault/README.md
```diff
@@ -364,6 +422,7 @@ The transaction creates an `AccountRoot` object for the `_pseudo-account_`. Ther
 - Create a new `MPTokenIssuance` ledger object for the vault shares.
   - If the `DomainID` is provided:
     - `MPTokenIssuance(Vault.MPTokenIssuanceID).DomainID = DomainID` (Set the Permissioned Domain ID).
+  - Create an `MPToken` object for the Vault Owner.
```
Bronek: I would clarify that this `MPToken` is to hold vault shares.

XLS-0065-single-asset-vault/README.md
```diff
@@ -164,22 +166,41 @@ The `Vault` object costs one reserve fee per object created:
...
 
-##### 2.1.6.1 `MPTokenIssuance`
+##### 2.1.6.1 `Scale`
+
+The **`Scale`** field enables the vault to accurately represent fractional asset values using integer-only MPT shares, which prevents the loss of value from decimal truncation. It defines a scaling factor, calculated as $10^{\text{Scale}}$, that converts a decimal asset amount into a corresponding whole number of shares. For example, with an `Scale` of `6`, a deposit of **20.3** assets is multiplied by $10^6$ and credited as **20,300,000** shares.
+
+As a general rule, all calculations involving MPTs are executed with a precision of a single MPT, treating them as indivisible units. If a calculation results in a fractional amount, it will be rounded up, down or to the nearest whole number depending on the context. Crucially, the rounding direction is determined by the protocol and is not controlled by the transaction submitter, which may lead to unexpected results for the fractional part.
```
Bronek: maybe remove the last part " for the fractional part" ? I think it is obvious.

XLS-0065-single-asset-vault/README.md
```diff
@@ -164,22 +166,41 @@ The `Vault` object costs one reserve fee per object created:
...
+
+##### 2.1.6.1.1 `IOU`
+
+When a vault holds an **`IOU`**, the `Scale` is configurable by the Vault Owner at the time of the vault's creation. The value can range from **0** to a maximum of **18**, with a default of **6**. This flexibility allows issuers to set a level of precision appropriate for their specific token.
+
+##### 2.1.6.1.2 `XRP`
+
+When a vault holds **`XRP`**, the `Scale` is fixed at **0**. This aligns with XRP's native structure, where one share represents one "drop" (the smallest unit of XRP), and one XRP equals 1,000,000 drops. Therefore, a deposit of 10 XRP to an empty Vault will result in the issuance of 10,000,000 shares ($10 \times 10^6$).
```
Bronek: would _maybe_ add (inside the parentheses) ", that is one share per one drop", but unsure about it. Up to you.

XLS-0065-single-asset-vault/README.md
```diff
@@ -164,22 +166,41 @@ The `Vault` object costs one reserve fee per object created:
 
 Shares represent the portion of the Vault assets a depositor owns. Vault Owners set the currency code of the share and whether the token is transferable during the vault's creation. These two values are immutable. The share is represented by a [Multi-Purpose Token](https://github.com/XRPLF/XRPL-Standards/tree/master/XLS-0033-multi-purpose-tokens). The MPT is issued by the vault's pseudo-account.
 
-##### 2.1.6.1 `MPTokenIssuance`
+##### 2.1.6.1 `Scale`
+
+The **`Scale`** field enables the vault to accurately represent fractional asset values using integer-only MPT shares, which prevents the loss of value from decimal truncation. It defines a scaling factor, calculated as $10^{\text{Scale}}$, that converts a decimal asset amount into a corresponding whole number of shares. For example, with an `Scale` of `6`, a deposit of **20.3** assets is multiplied by $10^6$ and credited as **20,300,000** shares.
```
Bronek: "an Scale" does not seem grammatically correct.

XLS-0065-single-asset-vault/README.md
Tapanito: I'm not certain that's necessary. The tangential clarification in that sentence is that a drop is the smallest unit of XRP. The main point of that sentence is that we maintain 1:1 ratio with drops, not XRP.

PR #375, opened 2025-09-17
modify MutableFlags naming

yinyiqian1:


XLS-0094-dynamic-MPT/README.md
```diff
@@ -60,17 +60,19 @@ Bits in `MutableFlags` indicate specific fields or flags may be modified after i
...
+| [Reserved]                   |  ️`0x0001`   |       1       | [Reserved; To align with `Flags` values, the `MutableFlags` value starts from `0x0002`.] |
+| `tmfMPTCanMutateCanLock`     |  ️`0x0002`   |       2       | Indicates flag `lsfMPTCanLock` can be changed                                            |
+| `tmfMPTCanMutateRequireAuth` |  ️`0x0004`   |       4       | Indicates flag `lsfMPTRequireAuth` can be changed                                        |
+| `tmfMPTCanMutateCanEscrow`   |   `0x0008`   |       8       | Indicates flag `lsfMPTCanEscrow` can be changed                                          |
+| `tmfMPTCanMutateCanTrade`    |   `0x0010`   |      16       | Indicates flag `lsfMPTCanTrade` can be changed                                           |
+| `tmfMPTCanMutateCanTransfer` |  ️`0x0020`   |      32       | Indicates flag `lsfMPTCanTransfer` can be changed                                        |
+| `tmfMPTCanMutateCanClawback` |  ️`0x0040`   |      64       | Indicates flag `lsfMPTCanClawback` can be changed                                        |
```
mDuo13: For clarity, the hex values should use the full bit size, e.g. `0x00000001` instead of `0x0001`.

mDuo13 (approved):
:+1:

Left one nitpick that can be addressed at your leisure.

XLS-0094-dynamic-MPT/README.md
yinyiqian1: I added the full bit. But I'm still running `npx prettier --write .` locally, which was stuck for a while. I'll update soon.

XLS-0094-dynamic-MPT/README.md
```diff
@@ -95,17 +97,19 @@ A new optional field, `MutableFlags` (SField `sfMutableFlags`), is added to the
...
+| `lmfMPTCanMutateRequireAuth` | ️`0x00000004` |       4       | Indicates flag `lsfMPTRequireAuth` can be changed                                            |
+| `lmfMPTCanMutateCanEscrow`   | `0x00000008`  |       8       | Indicates flag `lsfMPTCanEscrow` can be changed                                              |
+| `lmfMPTCanMutateCanTrade`    | `0x00000010`  |      16       | Indicates flag `lsfMPTCanTrade` can be changed                                               |
+| `lmfMPTCanMutateCanTransfer` | ️`0x00000020` |      32       | Indicates flag `lsfMPTCanTransfer` can be changed                                            |
+| `lmfMPTCanMutateCanClawback` | ️`0x00000040` |      64       | Indicates flag `lsfMPTCanClawback` can be changed                                            |
+| `lmfMPTCanMutateMetadata`    | `0x00010000`  |     65536     | Allows field `MPTokenMetadata` to be modified                                                |
+| `lmfMPTCanMutateTransferFee` | `0x00020000`  |    131072     | Allows field `TransferFee` to be modified                                                    |
```
sappenin: Minor nit, but shouldn't we retain the same naming convention for mutable ledger-state flags that we do for mutable transaction flags? For example, above we indicate a mutable `tf` flag my adding an `m` in front of the `f`, so the flag name becomes `tmf...`.

If we follow that same pattern for `lsf` flags, then we should add an `m` to an `lsf` flag in the same manner, yielding `lsmf...` (I read this as `Ledger State Mutable Flag`).

The only reason I'm thinking about this is the definition of an `lsf` flag, which is to indicate it's a Ledger State Flag. `lmf` doesn't easily connect to the `state` word.

This is a minor nit, so feel free to ignore, but figured I'd voice this because flags can be confusing, and the closer we keep the acronyms to actual words, the better.

yinyiqian1: @mDuo13 Do you have any preference?

yinyiqian1: @sappenin I just updated the naming.

sappenin (approved):
LGTM!

PR #392, opened 2025-10-09
Add templates for the new XLS format

mvadari:


XLS_TEMPLATE.md
```diff
@@ -20,15 +20,17 @@
   TODO: Remove this comment before submitting
 -->
```
Tapanito: The preamble here does not match the preamble in XLS-0001

PR #406, opened 2025-10-28
XLS-101: Smart Contracts

mvadari:


XLS-0101-smart-contracts/README.md
```diff
@@ -0,0 +1,965 @@
...
+| `ContractAccount`         | ✔️        | `string`  | `AccountID`         | The pseudo-account that hosts this contract.                                                                                   |
+| `Owner`                   | ✔️        | `string`  | `AccountID`         | The owner of the contract, which defaults to the account that deployed the contract.                                           |
+| `Flags`                   | ✔️        | `number`  | `UInt32`            | Flags that may be on this ledger entry.                                                                                        |
+| `Sequence`                | ✔️        | `string`  | `UInt16`            | The ledger entry's type (`ContractSource`).                                                                                    |
+| `ContractHash`            | ✔️        | `string`  | `Hash256`           | The hash of the contract's code.                                                                                               |
```
dangell7, line 185: Contract Hash or the hash of prefix + `ContractHash`?

We cannot link them in the explorer very easily if we do just wasm hash

dangell7: Or add ContractSourceID for easy linking?

XLS-0101-smart-contracts/README.md
```diff
@@ -0,0 +1,965 @@
...
+
+### 10.1. Fields
+
+| Field Name        | Required? | JSON Type | Internal Type | Description                                                                                                                                        |
+| :---------------- | :-------- | :-------- | :------------ | :------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `TransactionType` | ✔️        | `string`  | `UInt16`      | The transaction type (`ContractUserDelete`).                                                                                                       |
+| `Account`         | ✔️        | `string`  | `AccountID`   | The account sending the transaction (the user).                                                                                                    |
+| `ContractAccount` |           | `string`  | `AccountID`   | The pseudo-account hosting the contract that is to be changed. This field is not needed if the pseudo-account is emitting this transaction itself. |
```
dangell7, line 481: Does the contract really need to call this? Would a host api be better? Emitting a transaction to do this seems inefficient.

XLS-0101-smart-contracts/README.md
```diff
@@ -0,0 +1,965 @@
...
+| `Sequence`                | ✔️        | `string`  | `UInt16`            | The ledger entry's type (`ContractSource`).                                                                                    |
+| `ContractHash`            | ✔️        | `string`  | `Hash256`           | The hash of the contract's code.                                                                                               |
+| `InstanceParameterValues` |           | `array`   | `STParameterValues` | The parameters that are provided by this deployment of the contract.                                                           |
+| `URI`                     |           | `string`  | `Blob`              | A URI that points to the source code of this contract, to make it easier for tooling to find the source code for the contract. |
+
+_TODO: should the `URI` field live on `ContractSource` or `Contract`?_
+
+#### 4.1.1. Object ID
+
+hash of prefix + `ContractHash` + `Sequence`
```
dangell7, line 193: Note: Original ContractHash + Sequence. The actual contract hash may change through ContractModify transactions

XLS-0101-smart-contracts/README.md
dangell7: @mvadari

XLS-0101-smart-contracts/README.md
```diff
@@ -0,0 +1,965 @@
...
+- All parameter types must be valid XRPL `STypes` (maybe ban `STObjects` and `STArrays` and other complex STypes like `Transaction`)
+  - Advantage of storing the parameter types separately: you don’t have to open the VM to check if the parameters provided are valid (more efficient). You also get the ABI on-chain.
+  - TODO: Maybe all variables in the smart contract must be STypes?
+- No more than 4 parameters allowed (for now - we can relax that restriction later)
+- All parameters are required, no overloading allowed
+  - This could potentially change in a future version
+- Possible flags:
+  - `tfSendAmount` - if the type is an `STAmount`, then that amount will be sent to the contract from your funds
+  - `tfSendNFToken` - if the type is a `Hash256`, then the `NFToken` with that ID will be sent to the contract from your holdings
+  - `tfAuthorizeTokenHolding` - if the type is an `STIssue` or `STAmount`, then you can automatically create a trustline/MPToken for that token (assuming it’s not XRP).
```
dangell7, line 141: Doing this on ContractCall could create an exploit. If this is allowed then I could create 100's of trustlines on the contract. Either we limit to Create/Modify or the contract itself just authorizes the trustline through an init function. (Imo authorize in the contract is a better route)

mvadari: I don't quite follow - this is the _user_ authorizing holding a token, not the _contract_. The contract would be the issuer in this case. And you're only authorizing that one token.

XLS-0101-smart-contracts/README.md
mvadari: Yeah that's fair. Should we add this to this list of transaction types that can't be emitted, then?

XLS-0101-smart-contracts/README.md
mvadari: I was thinking the hash directly, but I don't think I have a strong opinion on this - so fine with `ContractSourceID`

XLS-0101-smart-contracts/README.md
mvadari: Should we go back to `Owner` + `Sequence`? Just needs to be something, right? Now that I think about it, `ContractHash` + `Sequence` could result in a hash collision if two accounts with the same sequence number create the same contract.

XLS-0101-smart-contracts/README.md
dangell7: If you want the contract to be able to hold a token, the contract needs to add the trust line. I think I realize now thats not what you meant by this?

You mean if the contract creates an IOU/MPT that requires auth then this authorizes the user? Why wouldn't the user just use SetTrust or MPTAuthorize?

XLS-0101-smart-contracts/README.md
dangell7: > Yeah that's fair. Should we add this to this list of transaction types that can't be emitted, then?

Yepp good point

XLS-0101-smart-contracts/README.md
dangell7: it is, in the code now its hash, owner, seq. I'm fine with owner + seq.

XLS-0101-smart-contracts/README.md
mvadari: > You mean if the contract creates an IOU/MPT that requires auth then this authorizes the user? Why wouldn't the user just use SetTrust or MPTAuthorize?

Yeah, the user automatically authorizes the token. Just saves you one step and makes it a bit more of a seamless experience is all, IMO. I figured there would be some need to "approve" holding a token, but it felt clunky to have that separate/just indicate a number of trustlines/MPTs you'd be willing to hold - this felt the cleanest to me.

dangell7: Ahh right, and we know that this contract uses an IOU/MPT and you must authorize first.

mvadari: Yeah - like an LP token.

PR #411, opened 2025-11-17
Updates VaultWithdraw failure conditions

Tapanito:
When withdrawing from a vault, the user can specify a destination account that will receive the assets. 

The implementation, does not create RippleState or MPToken objects if they do not exist, failing the transaction.

This change reflects this logic in the specification.

### DIFF
--- a/XLS-0085-token-escrow/README.md
+++ b/XLS-0085-token-escrow/README.md
@@ -80,7 +80,6 @@ The `EscrowCreate` transaction is modified as follows:
   - **MPTs**: The escrow `Amount` is deducted from the source's MPT balance. The `sfOutstandingBalance` of the MPT issuance remains unchanged. The `sfLockedAmount` is increased on both the source's MPT and the MPT issuance.
 - **Escrow Object Creation:**
   - The `Escrow` ledger object includes:
-    - `CancelAfter`: When the Escrow Expires (Required on IOU/MPT)
     - `Amount`: Tokens held in escrow.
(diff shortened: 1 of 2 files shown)


XLS-0065-single-asset-vault/README.md
```diff
@@ -613,7 +613,9 @@ In sections below assume the following variables:
     - The shares `MPToken.MPTAmount` of the `Account` is less than $\Delta_{share}$ (attempt to withdraw more shares than owned).
     - `Vault.AssetsAvailable` < `Amount` (the vault has insufficient assets).
 
-- The `Destination` account is specified and it does not have permission to receive the asset.
+- The `Destination` account is specified:
+  - The account does not have permission to receive the asset.
+  - The account does not have a `RippleState` or `MPToken` object for the asset.
```
ximinez, line 618: Would it be worth adding a case for "The account does not exist", or is it sufficient that it be implied by the other two checks. i.e. If the account doesn't exist, it certainly won't have a line or token.

ximinez (approved):

XLS-0065-single-asset-vault/README.md
```diff
@@ -8,7 +8,7 @@
   category: Amendment
   requires: [XLS-33](../XLS-0033-multi-purpose-tokens/README.md)
   created: 2024-04-12
-  updated: 2025-10-13
+  updated: 2025-11-17
```
mvadari, line 11: this should be updated

PR #484, opened 2026-02-26
Add PermissionedDomain to LoanBroker

Tapanito:

Introduces access control for the Lending Protocol via Permissioned Domains. A LoanBroker can be configured as private at creation time, restricting loan issuance to Borrowers with valid credentials.

Changes:
- Add `DomainID` field and `lsfLoanBrokerPrivate` flag to the `LoanBroker` ledger entry
- Add `tfLoanBrokerPrivate` transaction flag to `LoanBrokerSet`
- Add failure conditions for credential validation on `LoanSet`
- Add Access Control section (3.1.13) documenting the feature behavior
- Add invariants for the private flag and domain relationship
- The domain can be changed or cleared after creation; the private flag is immutable
- Credential checks do not apply to `LoanPay`, Borrowers can always repay

### Type of Change

- [ ] New XLS Draft


XLS-0066-lending-protocol/README.md
```diff
@@ -578,10 +603,17 @@ This transaction uses the standard transaction fee.
...
 
-##### 3.3.3.2 Protocol-Level Failures
+8. If `LoanBrokerID` is **not** specified (creating new):
+   1. `DomainID` is provided and is zero. (`temMALFORMED`)
+   2. `DomainID` is provided, and the `tfLoanBrokerPrivate` flag is not set. (`temINVALID`)
+
+9. If `LoanBrokerID` is specified (modifying existing):
+   1. `LoanBrokerID` is empty. (`temINVALID`)
+   2. Submitter is attempting to modify fixed fields (`ManagementFeeRate`, `CoverRateMinimum`, `CoverRateLiquidation`). (`temINVALID`)
+   3. Submitter is attempting to modify `tfLoanBrokerPrivate` flag on an existing `LoanBroker`. (`temINVALID`)
```
ximinez: Note that because of the nature of the `Flags` field, and the existence of universal flags, having the `tfLoanBrokerPrivate` field unset should be ignored - not treated as trying to change the flag.

Conversely, if the flag is set, that should always return as `temINVALID` regardless of the state of the `LoanBroker` object. (If it's already set, then the flag is redundant. If it's not set, then that's an attempted change.) At the implementation level, that means this check can be done in `preflight` - you don't need to know the current state of the `LoanBroker` object.

So I would suggest rephrasing this as something like
```
The `tfLoanBrokerPrivate` flag is set in the transaction
```

XLS-0066-lending-protocol/README.md
```diff
@@ -624,16 +659,29 @@ This transaction uses the standard transaction fee.
...
+   1. Set `LoanBroker.DomainID = DomainID` (Set the Permissioned Domain).
 
-**TBD**
+4. If `DomainID` is provided and is zero:
+   1. Clear `LoanBroker.DomainID` (Unset the Permissioned Domain).
+
+#### 3.3.6 Invariants
 
-#### 3.3.6 Example JSON
+1. If `LoanBroker.lsfLoanBrokerPrivate` flag is set, it cannot be unset.
```
ximinez, line 686: Info: Since there are no other ledger flags defined for the loan broker object, this means that the `Flags` field can not be changed, and so this can be checked in `NoModifiedUnmodifiableFields`.

XLS-0066-lending-protocol/README.md
```diff
@@ -1061,33 +1109,37 @@ The account specified in the `Account` field pays the transaction fee.
 14. `PaymentTotal <= 0`. (`temINVALID`)
 15. `PaymentInterval` is less than `60` seconds. (`temINVALID`)
 16. `GracePeriod` is less than `60` seconds or greater than the `PaymentInterval`. (`temINVALID`)
+17. The `Counterparty` field is not specified and the `CounterpartySignature` is not from the `LoanBroker.Owner`. (`temBAD_SIGNER`)
```
ximinez: This is a protocol-level check, because it requires loading the `LoanBroker.Owner`'s account root and/or signers list. For example, the owner may have a regular key defined. The transactor can't know that the regular key is allowed to sign for the owner without loading `AccountRoot(LoanBroker.Owner)` to get the `sfRegularKey` field.

If the signer(s) in the `CounterpartySignature` is/are not allowed, the failure code will be a `tef`, depending on which failure case is hit. e.g. `tefMASTER_DISABLED`, `tefBAD_AUTH`, `tefNOT_MULTI_SIGNING`, `tefBAD_SIGNATURE`, `tefBAD_QUORUM`. There may be others.

`temBAD_SIGNER` is only relevant to `LoanSet` if
1. it's a Batch inner tx and the`Counterparty` is not set, or
2. if it's a normal tx and there is no `CounterpartySignature` at all, or
3. if there is no `Counterparty` set and the `LoanBroker(LoanBrokerID)` doesn't exist on ledger.

XLS-0066-lending-protocol/README.md
```diff
@@ -1374,6 +1426,7 @@ This transaction uses the standard transaction fee.
 13. The payment is on-time and the `Amount` is less than the calculated `totalDue` for a periodic payment (`periodicPayment + loanServiceFee`). (`tecINSUFFICIENT_PAYMENT`)
 14. The `tfLoanFullPayment` flag is specified and `Loan.PaymentRemaining == 1` (use regular payment for the final payment). (`tecKILLED`)
 15. The `tfLoanFullPayment` flag is specified and the `Amount` is less than the calculated `totalDue` for a full early payment (`principalOutstanding + accruedInterest + prepaymentPenalty + ClosePaymentFee`). (`tecINSUFFICIENT_PAYMENT`)
+16. `LoanBroker.DomainID` is set and the borrower does not have valid credentials in the LoanBroker's `PermissionedDomain`. (`tecNO_AUTH` / `tecOBJECT_NOT_FOUND`)
```
ximinez: This contradicts several places where you've said that the borrower can always make a payment, regardless of credentials.

ximinez (changes requested):

XLS-0066-lending-protocol/README.md
Tapanito: Fixed. :facepalm:

XLS-0066-lending-protocol/README.md
Tapanito: Yep, I captured the other `temBAD_SIGNER` errors, and moved this particular failure down to protocol checks.

XLS-0066-lending-protocol/README.md
Tapanito: @a1q123456 fyi.

ximinez (approved):
All my concerns have been addressed.

XLS-0066-lending-protocol/README.md
```diff
@@ -590,19 +622,22 @@ This transaction uses the standard transaction fee.
...
 
-6. `LoanBroker` object with the specified `LoanBrokerID` does not exist on the ledger. (`tecNO_ENTRY`)
-7. The submitter `AccountRoot.Account != LoanBroker(LoanBrokerID).Owner`. (`tecNO_PERMISSION`)
-8. The transaction `VaultID` does not match `LoanBroker(LoanBrokerID).VaultID`. (`tecNO_PERMISSION`)
-9. `DebtMaximum` is being reduced to a non-zero value below the current `DebtTotal`. (`tecLIMIT_EXCEEDED`)
+1. `LoanBroker` object with the specified `LoanBrokerID` does not exist on the ledger. (`tecNO_ENTRY`)
+2. The submitter `AccountRoot.Account != LoanBroker(LoanBrokerID).Owner`. (`tecNO_PERMISSION`)
+3. The transaction `VaultID` does not match `LoanBroker(LoanBrokerID).VaultID`. (`tecNO_PERMISSION`)
+4. `DebtMaximum` is being reduced to a non-zero value below the current `DebtTotal`. (`tecLIMIT_EXCEEDED`)
+5. `DomainID` is provided, is non-zero and `LoanBroker.lsfLoanBrokerPrivate` flag is not set. (`tecNO_PERMISSION`)
```
a1q123456: do we allow the case where `LoanBroker.lsfLoanBrokerPrivate` flag is not set, and domainID is provided as zero?

Tapanito: Good question! I think if `LoanBroker.lsfLoanBrokerPrivate` flag is **not** set, and DomainID is provided in any form, the TX should fail with `tecNO_PERMISSION`. What do you think?

a1q123456: Makes sense! Thanks for clarifying!

tyalymov (approved):

whawk46:
This is a useful addition and the issuance-time semantics read clearly. One property
isn't stated, and because §3.1.13 names KYC as the motivating case I think it needs to be:
**what happens to an outstanding loan when the borrower stops satisfying the domain.**

The spec covers the domain being changed or cleared by the owner, and is explicit that
`LoanPay` is unaffected. It doesn't cover the borrower's side — credential revoked,
credential expired, or the domain deleted underneath them. On XRPL that isn't a policy
choice a protocol gets to make; it's a property of how permissioned domains enforce.

## Measured

I've been building an attested credential issuer on top of XLS-70/80 and ran the
revocation paths on Devnet on 2026-08-05. These probes are against the **permissioned
DEX**, not LoanBroker — LoanBroker isn't deployed anywhere I can reach — so treat this as
evidence about the *mechanism* #484 adopts rather than about LoanBroker itself.

Three different ways of ceasing to qualify, all identical in shape:

| what happened | new action | object already on the ledger |
|---|---|---|
| issuer swaps the accepted credential (revocation) | `tecNO_PERMISSION` | **survives** |
| **domain deleted entirely** (`tesSUCCESS`) | `tecNO_PERMISSION` | **survives** |
| credential `Expiration` passes | `tecNO_PERMISSION` | **survives** |

Enforcement is at action time and never retroactive. Revocation blocks the *next* action
immediately — which is good, and the part people expect — but nothing already on the
ledger is unwound. Deleting the whole domain succeeds and still leaves resting objects
untouched.

Applied to #484: a borrower who loses KYC mid-loan keeps the loan. I suspect that is the
*correct* design — you cannot unwind a disbursed loan, and forcing liquidation on a
credential change would be worse — but it is exactly the sentence a compliance officer
will need, and right now it has to be discovered rather than read.

Suggested addition to §3.1.13, roughly:

> Credential validation applies at loan issuance only. If a Borrower's credential is
> later revoked or expires, or the Permissioned Domain is deleted, existing Loans are
> unaffected and remain payable; only new `LoanSet` transactions are refused. The ledger
> does not unwind or liquidate outstanding Loans on credential change.

## Two smaller notes

**`tecEXPIRED` arrives silently.** Expiry is the one exit that no party triggers — there
is no transaction, no event, nothing on the ledger marking the moment a Borrower stops
qualifying. An issuer relying on expiry for periodic re-KYC has to poll for it. Worth a
line, since the other two exits are someone's deliberate act and this one isn't.

**The owner reserve transfers to the Borrower on accept.** Measured on the same run:
`CredentialCreate` puts the reserve on the issuer (issuer OwnerCount +1, subject +0), and
`CredentialAccept` moves it (issuer −1, subject +1). So a Borrower must have spare
reserve capacity before they can hold a KYC credential at all. For a lending flow that is
an onboarding step ahead of the loan, and it's easy to miss when reading §3.1.13 alone.

## Offer

Happy to run the same probes directly against LoanBroker once it's on a network I can
reach, and to publish the results the same way — including the case I'd most want checked
before merge: whether a `LoanSet` that is signed but not yet applied behaves like a
resting offer when the credential lapses in between.

---

*Disclosure: written with significant AI assistance, under human direction, review, and
correction.*

XLS-0066-lending-protocol/README.md
Tapanito: The specification now states this explicitly: a `DomainID` provided in any form, including zero, on a broker without `lsfLoanBrokerPrivate` fails with `tecNO_PERMISSION` (§3.4.4.2). That matches the implementation in the companion rippled PR, where the check tests field presence rather than a non-zero value, and its regression test asserts `tecNO_PERMISSION` for a zero `DomainID` on a public broker.

Tapanito:
Good catch, and you are right that this is enforcement behaviour rather than a policy choice. The borrower side is now stated in the spec.

Credential authorization is evaluated when a transaction is applied, against the ledger state at that time, and it is not re-checked against loans that already exist. So a revoked credential, an expired credential, or a deleted `PermissionedDomain` does not touch an outstanding `Loan`: repayment through `LoanPay` continues to work, and no existing loan becomes unenforceable. What those events do prevent is new borrowing — a subsequent `LoanSet` against that broker fails the domain check.

This matches the implementation: credentials are verified during application (`LoanSet.cpp:366-385,428-438`), `PermissionedDomainDelete` removes only the domain object and touches nothing that references it (`PermissionedDomainDelete.cpp:45-69`), and there is a regression test covering revocation leaving repayment intact while blocking new loans (`Loan_test.cpp:8946-8996`).

I left the general credential expiry-event and credential-reserve behaviour to the XLS-70 documentation rather than restating it here.

mvadari:
@ai-review --aggressive

XLS-0066-lending-protocol/README.md
Tapanito: Already matches develop: an unset `tfLoanBrokerPrivate` is ignored; setting it on modify returns `temINVALID` in preflight (`LoanBrokerSet.cpp`).

XLS-0066-lending-protocol/README.md
```diff
@@ -578,10 +606,19 @@ This transaction uses the standard transaction fee.
...
+8. The `LendingPermissionedDomain` amendment is not enabled and either `DomainID` is provided or the `tfLoanBrokerPrivate` flag is set. (`temDISABLED`)
+
+9. If `LoanBrokerID` is **not** specified (creating new):
+   1. `DomainID` is provided and is zero. (`temMALFORMED`)
+   2. `DomainID` is provided, and the `tfLoanBrokerPrivate` flag is not set. (`temINVALID`)
+
+10. If `LoanBrokerID` is specified (modifying existing):
+11. `LoanBrokerID` is empty. (`temINVALID`)
+12. Submitter is attempting to modify fixed fields (`ManagementFeeRate`, `CoverRateMinimum`, `CoverRateLiquidation`). (`temINVALID`)
+13. The `tfLoanBrokerPrivate` flag is set. (`temINVALID`)
```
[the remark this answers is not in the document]
Tapanito: The modify-only checks are nested under the existing-`LoanBrokerID` item so they do not apply to create.

XLS-0066-lending-protocol/README.md
```diff
@@ -1089,6 +1142,11 @@ The account specified in the `Account` field pays the transaction fee.
 23. The `LoanBroker.Owner` is not authorized for the asset. (`tecNO_AUTH`)
 24. The `LoanBroker.LoanSequence` has reached its maximum value. (`tecMAX_SEQUENCE_REACHED`)
 
+25. `LoanBroker.lsfLoanBrokerPrivate` flag is set:
+    1. `LoanBroker.DomainID` is not set (LoanBroker is private, but domain is not configured). (`tecNO_AUTH`)
+    2. The `PermissionedDomain` object referenced by `LoanBroker.DomainID` does not exist. (`tecOBJECT_NOT_FOUND`)
+    3. `Borrower` does not have credentials in the LoanBroker's `PermissionedDomain`. (`tecNO_AUTH` / `tecEXPIRED`)
```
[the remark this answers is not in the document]
Tapanito: Split into separate `tecNO_AUTH` and `tecEXPIRED` conditions, matching `validDomain` / `verifyValidDomain` in `CredentialHelpers.cpp` and `LoanSet` preclaim/doApply.

XLS-0066-lending-protocol/README.md
```diff
@@ -1513,6 +1571,8 @@ The management fee is calculated as a percentage of interest earned rather than
 
 The protocol makes strong trust assumptions between Vault Depositors, LoanBrokers, and Borrowers. The protocol does not offer on-chain algorithmic protection against default, thus all protocol participants must perform their due diligence and necessary off-chain checks.
 
+Private LoanBrokers additionally rely on the accepted credential issuers and the LoanBroker Owner's Permissioned Domain configuration. Credential checks provide point-in-time authorization for new Loans; revocation, expiration, or Domain deletion does not unwind an outstanding Loan or prevent repayment. LoanBroker Owners must account for that non-retroactive behavior when using credentials for compliance controls.
```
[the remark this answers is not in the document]
Tapanito: Documented that the referenced domain owner may differ from the LoanBroker Owner and can change `AcceptedCredentials` or delete the domain.

XLS-0066-lending-protocol/README.md
```diff
@@ -388,6 +399,16 @@ Lending Protocol:
 - CoverAvailable = CoverAvailable − DefaultCovered
   = 1,000 − 10.9 = **989.1 Tokens**
 
+#### 3.1.13 Access Control
+
+A Loan Issuer may want to ensure that all Borrowers meet certain legal obligations, such as KYC. This can be done by attaching a [Permissioned Domain](../XLS-0080-permissioned-domains/README.md), which restricts Loans to accounts with specific credentials. Only accounts with the required credentials can take out loans from a private LoanBroker. To configure a LoanBroker as private, set the `lsfLoanBrokerPrivate` flag when creating the `LoanBroker` object. If this flag is enabled, the LoanBroker is private. If the flag is not set, the LoanBroker is public, and any account may take out a loan.
```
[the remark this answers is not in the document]
Tapanito: Create/configure now names `tfLoanBrokerPrivate`; the resulting ledger flag is `lsfLoanBrokerPrivate`.

XLS-0066-lending-protocol/README.md
```diff
@@ -388,6 +399,16 @@ Lending Protocol:
...
+
+The presence or absence of a Permissioned Domain does not affect the signature requirements for the `LoanSet` transaction. Both the LoanBroker Owner and the Borrower must sign the `LoanSet` transaction, ensuring that even if an account is included in the Permissioned Domain, both parties must explicitly authorise the transaction.
+
+The LoanBroker Owner can clear or change the associated Permissioned Domain. If a private LoanBroker does not have a Permissioned Domain attached, loans cannot be issued until one is configured. Removing or changing the Permissioned Domain does not affect the `LoanPay` transaction for existing or new loans. Borrowers can always make loan payments. **Note**, the LoanBroker Owner does not need credentials in the Permissioned Domain to issue a loan to a Borrower. However, if the Owner wishes to take out a loan from their own LoanBroker, they must hold the required credentials.
```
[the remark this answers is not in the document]
Tapanito: Qualified: `LoanPay` has no domain credential check; repayment remains subject to the other `LoanPay` failure conditions.

PR #517, opened 2026-04-14
feat: add `get_function_hash` function to XLS-102

mvadari:

Title pretty much says it all

### Context of Change

A requested function

### Type of Change

## Future Tasks

Implementation work

### DIFF
--- a/XLS-0102-wasm-vm/README.md
+++ b/XLS-0102-wasm-vm/README.md
@@ -245,7 +245,7 @@ This section is the only section of functions that will likely be different for 
 | Function Signature                                                                         | Description                                                                                 | Gas Cost |
 | :----------------------------------------------------------------------------------------- | :------------------------------------------------------------------------------------------ | :------- |
 | `update_data(`<br/>&emsp;`data_ptr: i32,`<br/>&emsp;`data_len: i32`<br />`)`               | Update the `Data` field in the ledger object that hosts the WASM code, e.g. a Smart Escrow. | 50       |
-| `get_function_hash(`<br/>&emsp;`out_buff_ptr: i32,`<br/>&emsp;`out_buff_len: i32`<br />`)` | Get the hash of the current `FinishFunction`.                                               | 1000     |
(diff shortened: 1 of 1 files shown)


XLS-0102-wasm-vm/README.md
```diff
@@ -238,15 +238,14 @@ Each of these host functions will return `0` on success and a negative value on
...
-| :--------------------------------------------------------------------------- | :------------------------------------------------------------------------------------------ | :------- |
-| `update_data(`<br/>&emsp;`data_ptr: i32,`<br/>&emsp;`data_len: i32`<br />`)` | Update the `Data` field in the ledger object that hosts the WASM code, e.g. a Smart Escrow. | 50       |
+| Function Signature                                                                         | Description                                                                                 | Gas Cost |
+| :----------------------------------------------------------------------------------------- | :------------------------------------------------------------------------------------------ | :------- |
+| `update_data(`<br/>&emsp;`data_ptr: i32,`<br/>&emsp;`data_len: i32`<br />`)`               | Update the `Data` field in the ledger object that hosts the WASM code, e.g. a Smart Escrow. | 50       |
+| `get_function_hash(`<br/>&emsp;`out_buff_ptr: i32,`<br/>&emsp;`out_buff_len: i32`<br />`)` | Get the hash of the current `FinishFunction`.                                               | 35000    |
```
sappenin, line 248: Note for fututre: the engineering team realized in verbal discussions that the requirement here is not "get the hash of the currently running function" but instead "get the hash of a function in a different escrow."

The genesis of this was the system currently allows locating an Escrow object and then getting the field called `FinishFunction`. For large functions (e.g., 100k) this would be too resource intensive for our liking (contract would allocate a 100kb buffer, bytes would be loaded into it, then a hash computation, etc).

Instead, we decided to not allow getting all of the bytes of a FinishFunction, but then this begs the question - how to get the hash of a function living in a different Escrow.

One option (which nobody likes) is to use a virtual sfield (e.g., `FinishFunctionHash`) and let C++ code return a hash.

Another option (which nobody likes) is to store the hash in an escrow and create a real sfield for it. However, this means we have to store 34-ish extra bytes for every contract, even if nobody ever needs to get the hash.

mvadari:
We have decided to not implement this feature, due to lack of perceived utility.

PR #550, opened 2026-05-27
spec(XLS-65): Update VaultSet failure conditions and state changes

Tapanito:
Syncs the VaultSet spec section with the current implementation in `src/libxrpl/tx/transactors/vault/VaultSet.cpp`.

**Data Verification (3.3.2.1)** — was `_TBD_`, now populated:
- Empty or oversized Data field (`temMALFORMED`)
- Negative AssetsMaximum (`temMALFORMED`)
- No fields provided / nothing to update (`temMALFORMED`)

**Protocol-Level Failures (3.3.2.2)** — corrected and reduced to 5 accurate items:
- Removed item about immutable field modification (no such check in code)
- Removed item about nothing-to-update (moved to Data Verification)
- Corrected AssetsMaximum condition: only fails if the new value is non-zero and less than current AssetsTotal (removed spurious outer condition on current AssetsMaximum)

**State Changes (3.3.3)** — extended:
- Added: setting DomainID to zero removes it from the MPTokenIssuance (clearing the permissioned domain)


tyalymov (approved):
LGTM

mvadari (approved):

PR #551, opened 2026-05-27
spec(XLS-65): Update VaultDelete failure conditions and state changes

Tapanito:
Syncs the VaultDelete spec section with the current implementation in `src/libxrpl/tx/transactors/vault/VaultDelete.cpp`.

**Data Verification (3.4.2.1)** — was `_None._`, now has one item:

**Protocol-Level Failures (3.4.2.2)** — extended with error codes and split conditions:
- Split former item 3 into separate checks for `AssetsAvailable`, `AssetsTotal`, and `OutstandingAmount` (each returns `tecHAS_OBLIGATIONS` independently)
- Replaced vague item 4 (pseudo-account directory contents) with the actual runtime check: pseudo-account directory still contains objects after cleanup returns `tecHAS_OBLIGATIONS`

**State Changes (3.4.3)** — corrected ordering and extended:
- Added: vault owner's share MPToken is also deleted if it exists
- Corrected ordering to match code (asset holding removed first, then share MPToken, then MPTokenIssuance, then pseudo-account, then vault)
- Added: OwnerCount decremented by 2


tyalymov (approved):
LGTM

mvadari (approved):

PR #552, opened 2026-05-27
spec(XLS-65): Update VaultDeposit failure conditions and state changes

Tapanito:
Syncs the VaultDeposit spec section with the current implementation in `src/libxrpl/tx/transactors/vault/VaultDeposit.cpp`.

**Data Verification (3.5.2.1)** — was `_None._`, now has two items:
- Amount ≤ 0 (`temBAD_AMOUNT`)

**Protocol-Level Failures (3.5.2.2)** — extended from 7 to 13 items:
- Private vault auth: split into two sub-cases (no domain configured → `tecNO_AUTH`; not a valid domain member → `tecNO_AUTH`)
- Added: vault shares locked for depositor (`tecLOCKED`)
- Added: depositor lacks required authorized holding for MPT (`tecNO_AUTH`)
- Added: amount rounds to zero at vault scale (`tecPRECISION_LOSS`)
- Added: amount rounds to zero at depositor trust line scale / IOU (`tecPRECISION_LOSS`)
- Added: computed share amount is zero (`tecPRECISION_LOSS`)
- Added: arithmetic overflow in share calculation (`tecPATH_DRY`)

**State Changes (3.5.3)** — corrected:
- Use $\Delta_{asset}$ throughout (actual transferred amount may be less than requested `Amount` due to IOU scale rounding)
- Note on rounding added


tyalymov (commented):
Same question as in #549: in 3.5.2.2 the order of the failure conditions doesn't match the evaluation order in preclaim/doApply. Two spots:

private-domain auth is listed before the freeze/lock checks, but the code checks freeze/lock first;
tecINSUFFICIENT_FUNDS is listed before the vault-scale tecPRECISION_LOSS, but the code checks precision first.
Is the numbered list meant to be the evaluation order? If so, can we reorder to match?

Tapanito:
Confirmed against `VaultDeposit.cpp` — same answer as #549, the list is meant to be evaluation order. Two fixes: (1) moved the private-domain auth check to after the MPT-transferability/freeze/shares-locked checks; (2) moved the vault-scale `tecPRECISION_LOSS` check before `tecINSUFFICIENT_FUNDS` — the code rounds and rejects a zero amount before it ever checks the depositor's balance.

tyalymov (approved):

mvadari (approved):

PR #553, opened 2026-05-27
spec(XLS-65): Update VaultWithdraw failure conditions and state changes

Tapanito:
Syncs the VaultWithdraw spec section with the current implementation in `src/libxrpl/tx/transactors/vault/VaultWithdraw.cpp`.

**Data Verification (3.6.2.1)** — was `_None._`, now has three items:
- Amount ≤ 0 (`temBAD_AMOUNT`)
- Destination = 0 (`temMALFORMED`)

**Protocol-Level Failures (3.6.2.2)** — reworked:
- Removed MPT non-transferable check: post-`fixCleanup3_2_0`, VaultWithdraw waives `lsfMPTCanTransfer` (recovery path allowing withdrawal even when issuer revokes transferability)
- Added: vault shares frozen/locked for submitter (`tecFROZEN` / `tecLOCKED`)
- Added: destination lacks required authorization (`tecNO_AUTH`)
- Added: destination account conditions (`tecNO_DST`, `tecDST_TAG_NEEDED`, `tecNO_PERMISSION`)
- Added: computed share amount is zero (`tecPRECISION_LOSS`)
- Added: arithmetic overflow (`tecPATH_DRY`)

**State Changes (3.6.3)** — corrected:
- Fixed terminology throughout: "depositor" → "submitter" / "destination"
- Added: share MPToken is deleted when balance reaches zero (for non-owners)
- Clarified that "destination" is `Destination` if provided, otherwise `Account`
- Reordered to match code execution order


tyalymov (commented):
The definition changed from `Vault.AssetsTotal` to `the total assets in the vault, including any future yield`. 

Two things I'd check:

1. What field or formula is "future yield"? The withdrawal conversion in the code uses `Vault.AssetsTotal - Vault.LossUnrealized` with no yield term anywhere.
2. The rest of 3.6 uses concrete fields: State Changes and invariant 3.6.4 reconcile `Vault.AssetsTotal`, and the liquidity check uses `Vault.AssetsAvailable`. If `Gamma_asset` is a larger, yield-inclusive value, the conversion rate and the accounting no longer share a base, which could over-value shares on withdrawal.
Could we bind Gamma_asset to a concrete field , or spell out what "future yield" refers to?

Tapanito:
You're right — checked `VaultHelpers.cpp`: the withdraw conversion base is `Vault.AssetsTotal - Vault.LossUnrealized` (the ι term already used elsewhere in section 2.1.7), not a yield-inclusive figure. There's one wrinkle: when the withdrawer is the sole shareholder, the `LossUnrealized` subtraction is waived (they own 100% of both realized and unrealized value, so no proportional haircut applies). Rebound Γ_asset to `Vault.AssetsTotal - Vault.LossUnrealized`, noted the sole-shareholder waiver, and dropped "including any future yield" — it didn't correspond to anything in the implementation.

tyalymov (approved):

mvadari (approved):

PR #564, opened 2026-06-16
update: Changes to the XLS-68 PR before merge

mvadari:

Title says it all

### Context of Change

https://github.com/XRPLF/rippled/pull/7350

### Type of Change


XLS-0068-sponsored-fees-and-reserves/README.md
yinyiqian1: SponsoredOwnerCount might be sponsored by other accounts. So this calculation is not accurate

XLS-0068-sponsored-fees-and-reserves/README.md
```diff
@@ -910,17 +1002,19 @@ When a transaction includes a `Sponsor` field with the `spfSponsorFee` flag enab
...
 
-### 13.3. Permission Value
+This permission may not be used to submit any transactions (including `SponsorshipSet` or `SponsorshipTransfer`), only in the `SponsorSignature` field of another transaction.
+
+### 14.3. Permission Value
 
 65549
 
-## 14. Granular Permission: `SponsorReserve`
+## 15. Granular Permission: `SponsorReserve`
```
yinyiqian1: We currently do not support the granular permissions. We need to mention it

XLS-0068-sponsored-fees-and-reserves/README.md
```diff
@@ -470,46 +474,52 @@ The total fee calculation for signatures will now be $( 1+|tx.Signers| + |tx.Spo
...
+1. Only one or two of `Sponsor`, `SponsorFlags`, and `SponsorSignature` is included — they must either all be included, if the transaction is sponsored, or none, if it is not (`temMALFORMED`).
+1. `SponsorFlags` includes invalid flags — currently, the only two valid flags are `spfSponsorFee` and `spfSponsorReserve` (`temINVALID_FLAG`).
 
 #### 8.3.2. Fee Sponsorship Failures
 
-1. The sponsor's account does not have enough XRP to cover the sponsored transaction fee (`telINSUF_FEE_P`)
+1. The sponsor's account does not have enough XRP to cover the sponsored transaction fee (`terINSUF_FEE_B`, or `tecINSUFF_FEE` if some — but not enough — XRP is available when the transaction is applied to a closed ledger)
 
-If a `Sponsorship` object exists:
+If a `Sponsorship` object exists, it is **always** used as the source of the sponsored fee — even if a valid `SponsorSignature` is also included. A co-signature does not bypass the object's budget or limits:
```
oleks-rip, line 491: even if a valid `SponsorSignature` is also included  -> even if a valid `SponsorSignature` or `Delegate` is also included

XLS-0068-sponsored-fees-and-reserves/README.md
```diff
@@ -470,46 +474,52 @@ The total fee calculation for signatures will now be $( 1+|tx.Signers| + |tx.Spo
...
 
-1. The `lsfRequireSignatureForFee` flag is enabled and there is no sponsor signature included.
-1. There is not enough XRP in the `FeeAmount` to pay for the transaction.
-1. Paying fees via sponsorship will _not_ be able to [go below the reserve requirement](https://xrpl.org/docs/concepts/accounts/reserves#going-below-the-reserve-requirement).
-1. The fee in `tx.Fee` is greater than `Sponsorship.MaxFee`
+1. The `lsfSponsorshipRequireSignForFee` flag is enabled and there is no sponsor signature included (`terNO_PERMISSION`).
+2. There is not enough XRP in the `FeeAmount` to pay for the transaction (`terINSUF_FEE_B` / `tecINSUFF_FEE`). The transaction errors; it does **not** fall back to the sponsor's main `AccountRoot.Balance`, even if the sponsor co-signed the transaction.
```
oleks-rip, line 494: Fallback  + delegates  should be mentioned in 3.2 (copy or move)

XLS-0068-sponsored-fees-and-reserves/README.md
oleks-rip: I think this remark about priority should be moved / duplicated to 
3.2. The Sponsorship Flow (Not Pre-Funded)  Or 
3.3. The Sponsorship Flow (Pre-Funded)

And also + Delegate behavior

XLS-0068-sponsored-fees-and-reserves/README.md
```diff
@@ -470,46 +474,52 @@ The total fee calculation for signatures will now be $( 1+|tx.Signers| + |tx.Spo
...
 
 If a `Sponsorship` object does not exist:
 
-1. There is no sponsor signature included.
+1. There is no sponsor signature included (`terNO_PERMISSION`).
+
+In all cases:
+
+1. If the resolved fee payer does not have sufficient XRP available, the transaction errors (`terINSUF_FEE_B` / `tecINSUFF_FEE`).
+2. Paying a transaction fee via sponsorship — whether pre-funded from the `Sponsorship.FeeAmount` or co-signed from the sponsor's `AccountRoot.Balance` — will _not_ be able to [go below the reserve requirement](https://xrpl.org/docs/concepts/accounts/reserves#going-below-the-reserve-requirement) (`terINSUF_FEE_B` / `tecINSUFF_FEE`). This is unlike an ordinary, non-sponsored fee, which may draw the paying account below its reserve.
```
oleks-rip, line 504: I think this should be checked if it was implemented correctly in all places over the code.

XLS-0068-sponsored-fees-and-reserves/README.md
```diff
@@ -470,46 +474,52 @@ The total fee calculation for signatures will now be $( 1+|tx.Signers| + |tx.Spo
...
+_Note: if a transaction doesn't charge a fee (such as an account's first `SetRegularKey` transaction), the transaction will still succeed._
 
 #### 8.3.3. Reserve Sponsorship Failures
 
-1. The sponsor does not have enough XRP to cover the reserve (`tecINSUFFICIENT_RESERVE`)
-1. The transaction does not support reserve sponsorship (see section [8.3.4](#834-transactions-that-cannot-be-sponsored))
+1. The sponsor does not have enough XRP to cover the reserve (`tecINSUFFICIENT_RESERVE`).
+2. The transaction does not support reserve sponsorship — see [section 8.3.4](#834-transactions-that-cannot-be-sponsored) (`temMALFORMED`).
+3. The transaction creates a ledger object whose owner is an account other than `tx.Account` (e.g. `AMMClawback` creating a trust line on behalf of the holder) (`tecNO_SPONSOR_PERMISSION`).
+4. The transaction includes an `sfDelegate` field and `spfSponsorReserve` is enabled — reserve sponsorship combined with permissioned delegation is disallowed (`temINVALID`; see [section 18.1](#181-permissioned-delegation)).
```
oleks-rip: copy this to 3.3. too

XLS-0068-sponsored-fees-and-reserves/README.md
```diff
@@ -470,46 +474,52 @@ The total fee calculation for signatures will now be $( 1+|tx.Signers| + |tx.Spo
...
 
-If a `Sponsorship` object exists:
+If a `Sponsorship` object exists, it is **always** used for the sponsored reserve — even if a valid `SponsorSignature` is also included. Every sponsored reserve is checked against, and consumed from, the object's remaining `ReserveCount`; a co-signature does not bypass the budget:
```
oleks-rip: Should fee be paid for the signers if they are not used?

XLS-0068-sponsored-fees-and-reserves/README.md
```diff
@@ -582,18 +593,19 @@ This transaction uses the standard transaction fee (currently 10 drops, subject
...
 5. Both `tfSponsorshipSetRequireSignForReserve` and `tfSponsorshipClearRequireSignForReserve` are enabled (`temINVALID_FLAG`)
 6. `CounterpartySponsor` is specified (which means that the `Sponsee` is submitting the transaction) and `tfDeleteObject` is not enabled, as only the counterparty sponsor can create/update the `Sponsorship` object (`temMALFORMED`)
-7. `MaxFee` is less than the base fee or is not denominated in XRP (`temBAD_AMOUNT`)
-8. `MaxFee` is greater than `FeeAmount` (`temBAD_AMOUNT`)
-9. `FeeAmount` is not denominated in XRP (`temBAD_AMOUNT`)
-10. `CounterpartySponsor` or `Sponsee` does not exist on the ledger (`terNO_ACCOUNT`)
-11. `Owner == Sponsee` (attempting to create self-sponsorship) (`temMALFORMED`)
-12. `CounterpartySponsor` does not have sufficient XRP to cover the reserve for the `Sponsorship` object (`tecINSUFFICIENT_RESERVE`)
-13. If `tfDeleteObject` is enabled:
+7. `MaxFee` is not denominated in XRP (`temBAD_AMOUNT`)
```
oleks-rip: why   **`MaxFee` is less than the base**  condition removed?

XLS-0068-sponsored-fees-and-reserves/README.md
```diff
@@ -1113,9 +1209,40 @@ The response fields are nearly identical to `account_objects`.
 }
 ```
 
-## 17. Security Considerations
+## 18. Feature Interactions
+
+Sponsorship is a cross-cutting feature that touches almost every transaction type and interacts with several other XRPL features. This section is a catch-all for those interactions. The normative mechanics live in the per-feature sections above; this section consolidates the resolution rules so they can be reasoned about in one place.
+
+### 18.1. Permissioned Delegation
```
oleks-rip: I think this should be move to 3.2/3.3  to have all behavior in one place

XLS-0068-sponsored-fees-and-reserves/README.md
```diff
@@ -1307,16 +1434,18 @@ Sponsored transactions and reserves are a common feature across blockchain ecosy
...
 
 - **Co-signed sponsorship** gives sponsors fine-grained control over each transaction, ideal for high-value or sensitive operations.
 - **Pre-funded sponsorship** reduces operational overhead for sponsors who want to enable many transactions without being involved in each one, while still maintaining limits via `MaxFee` and `ReserveCount`.
 
-### 20.3. Other Designs Considered
+**Precedence: the `Sponsorship` object always governs.** If a `Sponsorship` object exists between the sponsor and sponsee, it is always the authoritative source for that sponsorship — its budgets (`FeeAmount`, `ReserveCount`) are the ones checked and consumed, and its limits (`MaxFee`) always apply, even when the sponsor also co-signs the transaction. A co-signature never bypasses the object or falls back to the sponsor's account balance. This keeps the accounting predictable: by creating the object, the sponsor has opted into explicit budgets, and those budgets are the single source of truth for what the sponsee may spend. A sponsor who wants to co-sign directly from their balance again can delete the `Sponsorship` object (or the relevant budget field) via `SponsorshipSet`.
```
oleks-rip: 22.2   -  this  explanation too - better to move to 3.2/3.3.  To much nuances to look all over the document

XLS-0068-sponsored-fees-and-reserves/README.md
mvadari: We shouldn't make the fee dependent on ledger state

XLS-0068-sponsored-fees-and-reserves/README.md
mvadari: Felt like an unnecessary check to add, given that a MaxFee lower than the base fee could be used to prevent any fee-based signatures.

XLS-0068-sponsored-fees-and-reserves/README.md
```diff
@@ -750,24 +766,25 @@ Additional failure conditions specific to `SponsorshipTransfer`:
...
    2. The `Sponsor` field or the `SponsorFlags` field is missing (`temMALFORMED`)
    3. The `SponsorFlags` field does not include the `spfSponsorReserve` flag (`temINVALID_FLAG`)
    4. The new sponsor account does not exist (`terNO_ACCOUNT`)
    5. The new sponsor does not have enough XRP to cover the reserve for this object/account (`tecINSUFFICIENT_RESERVE`)
+10. If transferring the sponsorship to a new sponsor:
+11. The transaction is not submitted by the sponsee (`tecNO_PERMISSION`)
+12. The `Sponsor` field or the `SponsorFlags` field is missing (`temMALFORMED`)
+13. The `SponsorFlags` field does not include the `spfSponsorReserve` flag (`temINVALID_FLAG`)
+14. The new sponsor account does not exist (`terNO_ACCOUNT`)
+15. The new sponsor does not have enough XRP to cover the reserve for this object/account (`tecINSUFFICIENT_RESERVE`)
```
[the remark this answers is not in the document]
mvadari: Fixed in a previous push — items 8 and 9 are now properly nested sub-lists under "If creating a new sponsorship" / "If transferring the sponsorship to a new sponsor" (see section 10.5).

XLS-0068-sponsored-fees-and-reserves/README.md
mvadari: Good catch — fixed in a previous push. Section 12.3 now reads: "the associated sponsors' `SponsoringOwnerCount` are decremented for each object they have sponsored that is owned by `tx.Account`", which handles the multi-sponsor case.

XLS-0068-sponsored-fees-and-reserves/README.md
mvadari: Addressed — the standalone `SponsorFee`/`SponsorReserve` granular permission sections were removed and replaced with section 14 ("Sponsor Authorization Mechanism"), which opens with an explicit note that those permissions (65549/65550) were proposed in an earlier revision but are **not** implemented, and that the actual mechanism does not use the delegation-permission system at all. Section 14.2 spells out the delegation interaction.

XLS-0068-sponsored-fees-and-reserves/README.md
mvadari: Folded this into section 3.6 ("Sponsorship Resolution Rules"), which now covers both the co-signature and `Delegate` precedence in one place — this text just links to it instead of restating it.

XLS-0068-sponsored-fees-and-reserves/README.md
mvadari: Rather than duplicating into 3.2, I pulled all the precedence/fallback/delegate rules into a new section 3.6 ("Sponsorship Resolution Rules") so there's one canonical place for this instead of scattered copies.

XLS-0068-sponsored-fees-and-reserves/README.md
mvadari: Same as the 3.2 comment above — moved to section 3.6 instead of duplicating across 3.2/3.3.

XLS-0068-sponsored-fees-and-reserves/README.md
mvadari: Consolidated into section 3.6 ("Sponsorship Resolution Rules") instead — same goal, one canonical section rather than 3.2/3.3.

XLS-0068-sponsored-fees-and-reserves/README.md
mvadari: This precedence explanation now lives in section 3.6 alongside the other resolution rules, so it's not spread across the doc anymore.

XLS-0068-sponsored-fees-and-reserves/README.md
mvadari: Fee paying is centralized in `Transactor.cpp` so only a couple places it needed to be processed.

yinyiqian1:
I ran cross AI scan through claude and gpt, would you mind checking those:

## 1. High severity — behavior does not match the implementation

| # | Spec § | Spec says | `develop` actually does | Where |
| --- | --- | --- | --- | --- |
| H1 | §8.4.2(1), §20.2, §23.2 | A sponsor named in `Sponsor` + `spfSponsorReserve` sponsors "any account/object created as part of the transaction"; §20.2 shows third-party account sponsorship this way | A new account's `Sponsor` is set **only** by `tfSponsorCreatedAccount`, and always to `tx.Account` (the payment sender). `tx.Sponsor` is never consulted for account creation, so §20.2's signed example yields `Sponsor = rOldB3E44…`, not `rSponsor1Vktvz…`. §11.2's own wording ("sponsored by the `tx.Account`") is the correct one | `src/libxrpl/tx/transactors/payment/Payment.cpp:503-521` |
| H2 | §8.3.4 | `Payment` is allow-listed for `spfSponsorReserve` "(account creation only)" | Exactly inverted: reserve sponsorship on a `Payment` covers objects **owned by `tx.Account`** (e.g. an auto-created `MPToken`) and never the created account, because the sponsor is only applied when the entry's owner is `tx.Account` | `src/libxrpl/ledger/helpers/SponsorHelpers.cpp:116-136` |
| H3 | §18.1, §A.6 | "A sponsor will never be stuck sponsoring… because it can submit a `SponsorshipTransfer` at any point"; §A.6 says the sponsee merely can't create new objects | `tfSponsorshipEnd` reserve-checks the **sponsee** and fails with `tecINSUFFICIENT_RESERVE` if the sponsee cannot carry the reserve. The sponsor genuinely can be stuck. §A.7 already describes the real behavior, contradicting §18.1/§A.6 | `src/libxrpl/tx/transactors/sponsor/SponsorshipTransfer.cpp:422-433` (objects), `:518-526` (accounts) |
| H4 | §10.5 | Failure list gives `tecINSUFFICIENT_RESERVE` only for the Create/Reassign paths (items 8.5, 9.5) | The End path can also fail with `tecINSUFFICIENT_RESERVE` (see H3); that condition is missing from the list entirely | same as H3 |
| H5 | §22.5, throughout | Behavior is gated on the single `Sponsor` amendment | Several behaviors are additionally gated on separate fix amendments: the object-level End reserve check requires `fixCleanup3_4_0`; the "sponsor only covers `tx.Account`'s own entries" assertion path differs pre-`fixCleanup3_2_0`; deleted-account cleanliness enforcement keys off `fixCleanup3_2_0 \|\| featureSponsor \|\| …`. None is mentioned | `SponsorshipTransfer.cpp:422`, `SponsorHelpers.cpp:119`, `src/libxrpl/tx/invariants/InvariantCheck.cpp:497` |
| H6 | §9.5 | On creation, a non-positive `RemainingOwnerCountDelta` "simply results in the field being omitted" | On creation, a **negative** `FeeAmountDelta` *or* `RemainingOwnerCountDelta` fails with `tecNO_PERMISSION` — even when the other field supplies a positive budget (e.g. `FeeAmountDelta = +1000000` with `RemainingOwnerCountDelta = -5` is rejected). Zero is already rejected earlier (`temBAD_AMOUNT` / `temINVALID`) | `SponsorshipSet.cpp:45-71`, `:196-198` |

---

## 2. Medium severity — wrong error codes, wrong magnitudes, misattributed enforcement

| # | Spec § | Spec says | `develop` actually does | Where |
| --- | --- | --- | --- | --- |
| M1 | §8.3.1(7) | Sponsorship fields on a transaction that doesn't support sponsorship → `temMALFORMED` | `temINVALID_FLAG`, which is what §8.3.4 itself says (the spec contradicts itself) | `src/libxrpl/tx/Transactor.cpp:208-216` |
| M2 | §8.3.1(6) | All three of `SigningPubKey`/`TxnSignature`/`Signers`, or other bad combinations → `temMALFORMED` | No `temMALFORMED` path exists. The combination is caught by the signature check ("Cannot both single- and multi-sign") and surfaces as an invalid-signature rejection | `src/libxrpl/protocol/STTx.cpp:425-426`, `:480-481` |
| M3 | §9.4(1) | `tx.Account` must equal `CounterpartySponsor` or `Sponsee`, else `temMALFORMED` | No such check, and it is structurally impossible: the unspecified side defaults to `tx.Account`. The real rule is that the *named* counterparty must differ from `tx.Account`, which is already item 10 | `SponsorshipSet.cpp:105-109` |
| M4 | §10.6(3) | Owner-count delta is "usually 1, but e.g. 2 for a legacy `SignerList` or `Vault`, type-dependent for `Oracle`" | A legacy (pre-`MultiSignReserve`) `SignerList` contributes **`2 + len(SignerEntries)`**, not 2. `Vault` and `Oracle` are not sponsorable at all, so their counts can never appear on this path | `SponsorHelpers.cpp:287-315`, `:266-285` |
| M5 | §10.5(7) | Supported `ObjectID` types = §4.3.1 list "except `Sponsorship`" | `AccountRoot` is also excluded — an `ObjectID` pointing at an `AccountRoot` gets `tecNO_PERMISSION` (account transfers must omit `ObjectID`). And `Sponsorship` was never in the §4.3.1 list, so the stated exception is confusing | `SponsorHelpers.cpp:266-285` |
| M6 | §4.3 (header), §4.3.2, §5.7, §6.3, §7.2 | These are presented as "Invariant Checks" | Only four sponsorship invariants exist (owner-count delta equality, per-entry `OwnerCount >= SponsoredOwnerCount`, sponsored-object owner-count delta cross-check, `SponsoringAccountCount` vs `sfSponsor`-presence delta), plus pseudo-account and deleted-account field checks. Everything else listed (allow-listed types, `Sponsor != owner`, `SponsoringOwnerCount > 0`, `lsfLowReserve` pairing, budget non-emptiness) is a transactor precondition, not an invariant | `src/libxrpl/tx/invariants/SponsorshipInvariant.cpp`, `InvariantCheck.cpp:547`, `:1069` |
| M7 | §6.3, §19.2, §23.3 | Global invariant `Σ SponsoredOwnerCount = Σ SponsoringOwnerCount`, and a test for it | No node ever computes a ledger-wide sum; only the per-transaction delta equality is checked. As written the test item is not implementable | `SponsorshipInvariant.cpp:94` |
| M8 | §17.2 | "Sponsored reserve checks happen **after** the fee is taken" | Not uniform: `SponsorshipTransfer` checks the sponsee's reserve against the **pre-fee** balance when the sponsee is `tx.Account`, and the post-fee balance otherwise | `SponsorshipTransfer.cpp:308-312` |
| M9 | §5.2 | `Flags` optional, default `0` | `sfFlags` is a required ledger-entry common field and is always written on creation | `src/libxrpl/protocol/LedgerFormats.cpp` (common fields), `SponsorshipSet.cpp:311` |
| M10 | §9.5 | Describes `MaxFee` only for the update path | Creation differs and is undocumented: `MaxFee: "0"` is silently dropped at creation (stored only when `> 0`), while on update `MaxFee: 0` clears the field. The spec also never states that an absent `MaxFee` means *uncapped* | `SponsorshipSet.cpp:299-300`, `Transactor.cpp:576-580` |

---

## 3. Low severity — examples and wording

| # | Spec § | Issue |
| --- | --- | --- |
| L1 | §20.1.2, §20.2.2, §20.3.2 | All three signed examples use `SponsorSignature.Signature`; the field is `TxnSignature` (`src/libxrpl/protocol/InnerObjectFormats.cpp:164-170`) |
| L2 | §13.6 | The inner `Payment` names `Sponsor` + `spfSponsorReserve` for a payment to an existing account and omits `tfSponsorCreatedAccount`, so per H1/H2 it sponsors nothing — the flagship `Batch` example demonstrates a no-op |
| L3 | §16.2 | `sponsored_objects` description is copy-pasted from `account_objects` ("entries in this account's owner directory… escrows where this account is the destination"), which describes owned, not sponsored, objects |
| L4 | §9.4(17) | Accurate, but the implementation's "no flags set" test is `getFlags() & tfUniversalMask`, so `tfFullyCanonicalSig` alone still counts as no flags (`SponsorshipSet.cpp:149-152`) |
| L5 | §5.6.2 | "No other fields may be specified in the deletion transaction" reads as forbidding `CounterpartySponsor`/`Sponsee`, but exactly one of them is still required |
| L6 | §14.1 | "Checked in this order" implies the `Sponsorship` object is consulted second; in fact a populated `SponsorSignature` makes the object's `lsfSponsorshipRequireSignFor*` flags irrelevant entirely (the budget still governs) (`Transactor.cpp:428-429`) |

---

mvadari:
Thanks for running these — really useful. Went through each item; a lot of them were already fixed in the pre-merge cleanup on this branch, and I pushed [9a6afa1](https://github.com/XRPLF/XRPL-Standards/pull/564/commits/9a6afa1) to address the remainder.

**Summary:** 7 items newly fixed (H1, H2, H3, M5, L4, L5, L6), 13 already addressed in the pre-merge cleanup, 1 partial (H5), 1 n/a (L2). Per-item notes below.

<details>
<summary><b>High severity</b></summary>

| # | Status | Notes |
|---|---|---|
| **H1** | ✅ fixed | Rewrote §8.4.2(1) to make clear that a reserve sponsor's `Sponsor` field only lands on **non-account** entries owned by `tx.Account`; account creation is a separate mechanism (`tfSponsorCreatedAccount`, always by `tx.Account`, never by `tx.Sponsor`). Also fixed a stale test bullet in §23.2. The §20.2 example was already updated pre-merge to sponsor only the fee (`spfSponsorFee`) with the account-reserve carve-out from `tfSponsorCreatedAccount`, so its resolved `Sponsor` on the new account is `tx.Account`, matching the impl. |
| **H2** | ✅ fixed | You're right, "no effect on `Payment`" was too strong. Rewrote §8.3.4 to spell out the actual carve-out: `spfSponsorReserve` on a `Payment` never sponsors the created account, but does still apply to non-account entries auto-created for `tx.Account` (e.g. an auto-created `MPToken`). |
| **H3** | ✅ fixed | §18.1 was already reworded pre-merge to acknowledge the End-path `tecINSUFFICIENT_RESERVE` and point at §A.7 for the workaround. §A.6 was still saying the sponsee "just can't create more objects", which understated it — updated to mention the failure explicitly and link to §A.7. |
| **H4** | ✅ already addressed | §10.5 item 10 (added pre-merge) covers the End-path `tecINSUFFICIENT_RESERVE`, including the `fixCleanup3_4_0` gate for the object-level variant. |
| **H5** | 🟡 partial | `fixCleanup3_4_0` is already called out in §22.5 and §10.5(10) as the one user-visible gate. I intentionally didn't add `fixCleanup3_2_0` to the spec — that one only guards internal invariant/assertion paths and doesn't produce a different externally-visible outcome, so putting it in the XLS felt like it'd add noise more than clarity. Happy to add it if you disagree. |
| **H6** | ✅ already addressed | §9.4(13) and §9.5 were already updated pre-merge to state the "each field checked independently" behavior — a negative `FeeAmountDelta` or `RemainingOwnerCountDelta` on creation fails outright even if the other field is valid. |

</details>

<details>
<summary><b>Medium severity</b></summary>

| # | Status | Notes |
|---|---|---|
| **M1** | ✅ already addressed | §8.3.1(7) was updated pre-merge to `temINVALID_FLAG`. |
| **M2** | ✅ already addressed | §8.3.1(6) was rewritten pre-merge to acknowledge there's no dedicated check — it falls out of the single-vs-multi-sign logic as `temINVALID`. |
| **M3** | ✅ already addressed | The stale `tx.Account == CounterpartySponsor or Sponsee` rule was removed pre-merge; the actual behavior (both defaulting, and named counterparty must differ) is what's documented now. |
| **M4** | ✅ already addressed | §10.6 already says the SignerList delta is `2 + len(SignerEntries)`. Since `Vault`/`Oracle` aren't in the §4.3.1 allow-list they can't reach this path, so I dropped the misleading examples. |
| **M5** | ✅ fixed | Reworded §10.5(7) to lead with the `AccountRoot`-omitted-`ObjectID` carve-out and drop the confusing "except `Sponsorship`" phrasing (`Sponsorship` was never in §4.3.1 to begin with). |
| **M6** | ✅ already addressed | §4.3 has a note added pre-merge explaining that "Invariant Checks" sections mix runtime invariants and per-transactor preconditions. |
| **M7** | ✅ already addressed | §6.3, §19.2, and §23.3 all say per-transaction delta equality rather than a ledger-wide sum. |
| **M8** | ✅ already addressed | §17.2 was rewritten pre-merge to acknowledge the pre-fee/post-fee inconsistency, with `SponsorshipTransfer` End-path called out specifically. |
| **M9** | ✅ already addressed | §5.2 marks `Flags` as required with default `0` and cross-references the common-fields rule in §4.1. |
| **M10** | ✅ already addressed | §9.5 documents `MaxFee: 0` being dropped at creation and absent-`MaxFee` meaning uncapped. |

</details>

<details>
<summary><b>Low severity</b></summary>

| # | Status | Notes |
|---|---|---|
| **L1** | ✅ already addressed | All three signed examples use `TxnSignature`, matching `InnerObjectFormats.cpp`. |
| **L2** | ➖ n/a | The §13.6 inner is a `CheckCreate`, not a `Payment` — a `Check` is in the §4.3.1 allow-list, so `spfSponsorReserve` does sponsor the created object. Not a no-op. |
| **L3** | ✅ already addressed | §16.2's `sponsored_objects` description was rewritten pre-merge to describe entries the account is the sponsor of. |
| **L4** | ✅ fixed | Added a clarifier that "no flags set" refers only to `SponsorshipSet`-specific flags; universal flags like `tfFullyCanonicalSig` are ignored. |
| **L5** | ✅ fixed | Reworded §5.6.2 so it's clear one of `CounterpartySponsor` or `Sponsee` is still required at deletion (to identify the target); "no other fields" only referred to budget/limit/flag fields. |
| **L6** | ✅ fixed | Rewrote §14.1 to spell out that a populated co-signature makes the object's `lsfSponsorshipRequireSignFor*` flags moot but the object's **budget** still governs, with a cross-link to §3.6 rule 1. |

</details>

Let me know if any of these still feel off.

XLS-0068-sponsored-fees-and-reserves/README.md
```diff
@@ -4,11 +4,10 @@
   description: Allow an account to fund fees and reserves on behalf of another account
   author: Mayukha Vadari (@mvadari)
   category: Amendment
-  status: Draft
+  status: Final
```
yinyiqian1, line 8: need to add implementation link as well

XLS-0068-sponsored-fees-and-reserves/README.md
```diff
@@ -470,59 +487,74 @@ The total fee calculation for signatures will now be $( 1+|tx.Signers| + |tx.Spo
...
-1. `SponsorSignature.TxnSignature` is invalid.
-1. `SponsorSignature.Signers` is invalid (the signer list isn't on the account, quorum isn't reached, the public key(s) are invalid, or signature(s) are invalid).
-1. `SponsorSignature.SigningPubKey` is invalid (the public key doesn't match the account's master key or regular key, or the public key is otherwise invalid).
-1. The `Sponsor` doesn't exist on the ledger.
-1. An invalid sponsorship flag is used.
-1. `SponsorSignature.SigningPubKey`, `SponsorSignature.TxnSignature`, and `SponsorSignature.Signers` are all included (or other incorrect combinations of signing fields).
-1. `Sponsor`, `SponsorFlags`, or `SponsorSignature` is included in a transaction that does not support sponsorship (see section [8.3.4](#834-transactions-that-cannot-be-sponsored)).
-1. Only one or two of `Sponsor`, `SponsorFlags`, and `SponsorSignature` is included (they must either all be included, if the transaction is sponsored, or none, if it is not).
-1. `SponsorFlags` includes invalid flags (currently, the only two valid flags are `spfSponsorFee` and `spfSponsorReserve`).
+1. `SponsorSignature.TxnSignature` is invalid (`temBAD_SIGNATURE`).
```
yinyiqian1, line 499: The numbers are all 1 in this section

XLS-0068-sponsored-fees-and-reserves/README.md
```diff
@@ -536,13 +568,12 @@ If a `Sponsorship` object does not exist, the `tx.Fee` value is decremented from
...
-
-The sponsor's `SponsoringOwnerCount` field will be incremented by the number of objects that are sponsored as a part of the transaction, and the `SponsoringAccountCount` field will be incremented by the number of new accounts that are sponsored as a part of the transaction.
-
-The sponsee's `SponsoredOwnerCount` field will be incremented by the number of objects that are sponsored as a part of the transaction.
-
-The `SponsoredOwnerCount`, `SponsoringOwnerCount`, and `SponsoringAccountCount` fields will be decremented when those objects/accounts are deleted.
+1. Any non-account ledger entry created as part of the transaction whose owner is `tx.Account` will have a `Sponsor` field pointing at `tx.Sponsor`. **Account** creation is a separate mechanism: a new `AccountRoot` is only sponsored when the creating `Payment` sets `tfSponsorCreatedAccount` (see [section 11.2](#112-flags)), and its `Sponsor` is unconditionally `tx.Account`, not `tx.Sponsor`. There is no third-party `spfSponsorReserve` path for account creation; `spfSponsorReserve` on a `Payment` only affects objects created for `tx.Account` (see [section 8.3.4](#834-transactions-that-cannot-be-sponsored)).
```
yinyiqian1, line 584: The numbers are all 1 in this section

XLS-0068-sponsored-fees-and-reserves/README.md
```diff
@@ -744,44 +796,51 @@ This transaction uses the standard transaction fee (currently 10 drops, subject
...
 
 ### 10.6. State Changes
 
-- The `Sponsor` field on the object specified by `ObjectID` is deleted if the `tx.Sponsor` is the object's `Owner`, otherwise the `Sponsor` field is updated to the new `tx.Sponsor`.
-- The old sponsor (if applicable) has its `SponsoringOwnerCount`/`SponsoringAccountCount` decremented by one.
-- The new sponsor (if applicable) has its `SponsoringOwnerCount`/`SponsoringAccountCount` incremented by one.
-- If there is no new sponsor, then the owner's `SponsoredOwnerCount` will be decremented by one.
+The state change applied to the `Sponsor` field is driven purely by the transaction's flag, not by comparing `tx.Sponsor` to the object's current owner (since `Sponsor` is never present on the `tfSponsorshipEnd` path — see the failure conditions above):
+
+1. `tfSponsorshipEnd`: the `Sponsor` field on the object specified by `ObjectID` (or on the sponsored `AccountRoot`) is unconditionally removed.
```
yinyiqian1, line 844: The numbers are all 1 here

XLS-0068-sponsored-fees-and-reserves/README.md
```diff
@@ -828,8 +887,9 @@ Additional failure conditions when `tfSponsorCreatedAccount` is enabled:
 
 1. `tfNoRippleDirect`, `tfPartialPayment`, or `tfLimitQuality` are enabled (`temINVALID_FLAG`)
 1. `Amount` specifies a non-XRP currency (`temBAD_AMOUNT`)
+1. `SendMax` or `Paths` is included (`temINVALID`)
```
yinyiqian1, line 903: The numbers are all 1 here

XLS-0068-sponsored-fees-and-reserves/README.md
```diff
@@ -442,7 +459,7 @@ There are two flag values that are supported:
 | `spfSponsorFee`     | `0x00000001` | Sponsoring (paying for) the fee of the transaction.                |
 | `spfSponsorReserve` | `0x00000002` | Sponsoring the reserve for any objects created in the transaction. |
 
-#### 8.1.2 `SponsorSignature`
+#### 8.1.2. `SponsorSignature`
```
yinyiqian1: do we need to mention the SPN and SPM prefix here

XLS-0068-sponsored-fees-and-reserves/README.md
mvadari: Markdown processes all the 1s to be a list, easier to keep updated that way

XLS-0068-sponsored-fees-and-reserves/README.md
mvadari: Markdown processes all the 1s to be a list, easier to keep updated that way

XLS-0068-sponsored-fees-and-reserves/README.md
mvadari: Markdown processes all the 1s to be a list, easier to keep updated that way

XLS-0068-sponsored-fees-and-reserves/README.md
mvadari: Markdown processes all the 1s to be a list, easier to keep updated that way

XLS-0068-sponsored-fees-and-reserves/README.md
mvadari: Done in 9b24cd7

XLS-0068-sponsored-fees-and-reserves/README.md
```diff
@@ -442,7 +460,7 @@ There are two flag values that are supported:
...
```
yinyiqian1, line 460: do we need to mention the SPN and SPM prefix here

XLS-0068-sponsored-fees-and-reserves/README.md
```diff
@@ -703,10 +755,11 @@ This scenario ends the sponsorship for a sponsored ledger object or account. The
 
 The following fields are used in this scenario:
 
-- `ObjectID` must be included (if sponsored object)
+- `ObjectID` must be included (if ending sponsorship of an object)
+- `Sponsee` must be included if the **sponsor** is submitting the transaction to end an _account_ sponsorship of another account (the target account is identified by `Sponsee`). If the **sponsee** is ending their own sponsorship, `Sponsee` must be omitted and the target is `tx.Account`.
```
yinyiqian1: Here it says only for account, this also applies to object, right?

XLS-0068-sponsored-fees-and-reserves/README.md
yinyiqian1: It also blocks sponsee's deletion, it is saved under sponsee's list as well. tecHAS_OBLIGATIONS

yinyiqian1, line 1787: Also need to mention sponsee needs to do so as well

XLS-0068-sponsored-fees-and-reserves/README.md
```diff
@@ -470,59 +488,74 @@ The total fee calculation for signatures will now be $( 1+|tx.Signers| + |tx.Spo
...
+2. The transaction does not support reserve sponsorship — see [section 8.3.4](#834-transactions-that-cannot-be-sponsored) (`temINVALID_FLAG`).
+3. The transaction includes an `sfDelegate` field and `spfSponsorReserve` is enabled (reserve sponsorship combined with permissioned delegation is disallowed, see [section 17.1](#171-permissioned-delegation)) (`temINVALID`).
 
-If a `Sponsorship` object exists:
+If a `Sponsorship` object exists (it is [always used if it exists](#36-sponsorship-resolution-rules)):
 
-1. The `lsfRequireSignatureForReserve` flag is enabled and there is no sponsor signature included.
-1. There is not enough remaining count in the `ReserveCount` to pay for the transaction.
+1. The `lsfSponsorshipRequireSignForReserve` flag is enabled and there is no sponsor signature included (`terNO_PERMISSION`).
+2. There is not enough remaining count in the `RemainingOwnerCount` to pay for the transaction (`tecINSUFFICIENT_RESERVE`), even if the sponsor co-signed the transaction.
```
yinyiqian1: TrustSet returns `tecINSUF_RESERVE_LINE`
And some other places, for example EscrowFinish, CheckCash, etc. returns `tecINSUF_RESERVE_LINE`

XLS-0068-sponsored-fees-and-reserves/README.md
```diff
@@ -470,59 +488,74 @@ The total fee calculation for signatures will now be $( 1+|tx.Signers| + |tx.Spo
...
-1. `SponsorFlags` includes invalid flags (currently, the only two valid flags are `spfSponsorFee` and `spfSponsorReserve`).
+1. `SponsorSignature.TxnSignature` is invalid (`temBAD_SIGNATURE`).
+1. `SponsorSignature.Signers` is invalid — the signer list isn't on the account, quorum isn't reached, the public key(s) are invalid, or signature(s) are invalid (`tefBAD_AUTH` / `tefBAD_QUORUM` / `tefBAD_SIGNATURE` as appropriate).
+1. `SponsorSignature.SigningPubKey` is invalid — the public key doesn't match the account's master key or regular key, or the public key is otherwise invalid (`tefBAD_AUTH` or `temBAD_SIGNATURE`).
+1. The `Sponsor` doesn't exist on the ledger (`terNO_ACCOUNT`).
+1. An invalid sponsorship flag is used (`temINVALID_FLAG`).
+1. `SponsorSignature.SigningPubKey`, `SponsorSignature.TxnSignature`, and `SponsorSignature.Signers` are all included, or other incorrect combinations of signing fields — this is not a dedicated check; it is caught by the same signature-validity logic used elsewhere ("cannot both single- and multi-sign"), so it surfaces as `temINVALID`, not a distinct malformed error.
```
yinyiqian1: temBAD_SIGNATURE is returned. The rewrite of the error code in preflight2 to temINVALID is unreachable.

XLS-0068-sponsored-fees-and-reserves/README.md
mvadari: Added a new §8.1.2.2 documenting the SPN (single-sign) / SPM (multi-sign) hash prefixes for `SponsorSignature`, gated on `fixCleanup3_4_0`, with a link to the CounterpartySignature counterpart in XLS-66. Done in 3d92063.

XLS-0068-sponsored-fees-and-reserves/README.md
mvadari: Good catch — clarified in §10.3.1 that `Sponsee` is only used for account-level ends; for object-level ends `ObjectID` alone identifies the target (via its `Owner` field). Done in 3d92063.

XLS-0068-sponsored-fees-and-reserves/README.md
mvadari: Correct — updated §5.6.3 to state that `Sponsorship` is a deletion blocker for **both** the sponsor and the sponsee (it's in both owner directories via `OwnerNode`/`SponseeNode`), and returns `tecHAS_OBLIGATIONS` on either side. Done in 3d92063.

mvadari: Added a paragraph in A.3 noting the sponsee must also clear any `Sponsorship` objects (where they're the counterparty) before deleting, with a cross-link to the updated §5.6.3. Done in 3d92063.

XLS-0068-sponsored-fees-and-reserves/README.md
mvadari: Updated §8.3.3 items 1 and 2 to note that the error code is `tecINSUFFICIENT_RESERVE` generally, or the transaction-type-specific variant when applicable — e.g. `tecINSUF_RESERVE_LINE` for `TrustSet` and other transactions that establish a trust line as a side effect (`EscrowFinish`, `CheckCash`). Done in 3d92063.

XLS-0068-sponsored-fees-and-reserves/README.md
mvadari: Fixed — §8.3.1(6) now says `temBAD_SIGNATURE` instead of `temINVALID`. Done in 3d92063.

XLS-0068-sponsored-fees-and-reserves/README.md
```diff
@@ -756,7 +764,7 @@ This scenario ends the sponsorship for a sponsored ledger object or account. The
 The following fields are used in this scenario:
 
 - `ObjectID` must be included (if ending sponsorship of an object)
-- `Sponsee` must be included if the **sponsor** is submitting the transaction to end an _account_ sponsorship of another account (the target account is identified by `Sponsee`). If the **sponsee** is ending their own sponsorship, `Sponsee` must be omitted and the target is `tx.Account`.
+- `Sponsee` is only used for **account-level** ends where the **sponsor** is submitting the transaction against another account; in that case, `Sponsee` names the target account. If the **sponsee** is ending their own account sponsorship, `Sponsee` must be omitted and the target is `tx.Account`. For **object-level** ends, `Sponsee` must be omitted regardless of who is submitting: `ObjectID` alone identifies both the object and (via its `Owner` field) the sponsee.
```
yinyiqian1: I think object-level also requires `Sponsee`. The code asks for `Sponsee` field and will check it really owns the object through `isLedgerEntryOwner`

XLS-0068-sponsored-fees-and-reserves/README.md
```diff
@@ -519,14 +527,14 @@ _Note: if a transaction doesn't charge a fee (such as an account's first `SetReg
 
 #### 8.3.3. Reserve Sponsorship Failures
 
-1. The sponsor does not have enough XRP to cover the reserve (`tecINSUFFICIENT_RESERVE`).
+1. The sponsor does not have enough XRP to cover the reserve (`tecINSUFFICIENT_RESERVE`, or the transaction-type-specific variant when applicable — e.g. `tecINSUF_RESERVE_LINE` for `TrustSet` and for other transactions that can establish a trust line as a side effect, such as `EscrowFinish` or `CheckCash`).
```
yinyiqian1: With the help of AI: 

Path | Code
-- | --
TrustSet, low/high reserve-set paths (TrustSet.cpp:537,579,626) | tecINSUF_RESERVE_LINE
TrustSet, line-creation path (TrustSet.cpp:697,707) | tecNO_LINE_INSUF_RESERVE
CheckCash, IOU (CheckCash.cpp:462) | tecNO_LINE_INSUF_RESERVE
EscrowFinish / EscrowCancel, IOU (EscrowHelpers.h:84) | tecNO_LINE_INSUF_RESERVE
MPT counterparts of all three | tecINSUFFICIENT_RESERVE
SponsorshipSet (SponsorshipSet.cpp:279,391) | tecUNFUNDED, already documented in §9.4

yinyiqian1 (approved):

XLS-0068-sponsored-fees-and-reserves/README.md
mvadari: Good catch — you are right, my previous fix was wrong. Verified against [`SponsorshipTransfer.cpp`](https://github.com/XRPLF/rippled/blob/develop/src/libxrpl/tx/transactors/sponsor/SponsorshipTransfer.cpp): for object-level ends, `sponseeID = tx.Sponsee.value_or(tx.Account)`, and `isLedgerEntryOwner(view, objectSle, sponseeID)` is enforced (returns `tecNO_PERMISSION` otherwise). So `Sponsee` is required whenever the sponsor is submitting against a target owned by another account — for both account- and object-level ends.

Reworded §10.3.1 in [573f370](https://github.com/XRPLF/XRPL-Standards/commit/573f3705d7ef54b91b0e15e51f008f16fed8f56c) to reflect this (single rule covering both levels, plus the `isLedgerEntryOwner` check).

XLS-0068-sponsored-fees-and-reserves/README.md
mvadari: Verified against rippled — your breakdown is accurate. `checkReserve` [defaults to `tecINSUFFICIENT_RESERVE`](https://github.com/XRPLF/rippled/blob/develop/include/xrpl/ledger/helpers/AccountRootHelpers.h#L106) and callers pass the more specific code as the last arg.

Replaced the too-terse `tecINSUF_RESERVE_LINE` catch-all in §8.3.3 with a path-by-path breakdown in [573f370](https://github.com/XRPLF/XRPL-Standards/commit/573f3705d7ef54b91b0e15e51f008f16fed8f56c):

- `tecINSUFFICIENT_RESERVE` — default (MPT, generic object creation)
- `tecINSUF_RESERVE_LINE` — `TrustSet` low/high reserve-set on existing lines
- `tecNO_LINE_INSUF_RESERVE` — line-creation paths (`TrustSet` line-creation, `CheckCash` IOU, `EscrowFinish`/`Cancel` IOU)
- `tecUNFUNDED` — `SponsorshipSet` (already documented in §9.4 items 10–11)

XLS-0068-sponsored-fees-and-reserves/README.md
```diff
@@ -195,19 +210,19 @@ This ensures that there can be at most one `Sponsorship` object per sponsor-spon
...
+| `OwnerNode`           | ✔️        | ✔️        | N/A           | `string`  | `UInt64`      | A hint indicating which page of the sponsor's owner directory links to this object, in case the directory consists of multiple pages.                                         |
+| `SponseeNode`         | ✔️        | ✔️        | N/A           | `string`  | `UInt64`      | A hint indicating which page of the sponsee's owner directory links to this object, in case the directory consists of multiple pages.                                         |
+| `PreviousTxnID`       |           | ✔️        | N/A           | `string`  | `Hash256`     | The identifying hash of the transaction that most recently modified this entry.                                                                                               |
+| `PreviousTxnLgrSeq`   |           | ✔️        | N/A           | `number`  | `UInt32`      | The ledger index that contains the transaction that most recently modified this object.                                                                                       |
```
Tapanito: Please don't use emojies. The columns take `Yes`, `No`, `Conditional` values. This comment applies to the whole specification. Please review all the tables.

XLS-0068-sponsored-fees-and-reserves/README.md
```diff
@@ -470,59 +496,78 @@ The total fee calculation for signatures will now be $( 1+|tx.Signers| + |tx.Spo
...
+1. An invalid sponsorship flag is used (`temINVALID_FLAG`).
+1. `SponsorSignature.SigningPubKey`, `SponsorSignature.TxnSignature`, and `SponsorSignature.Signers` are all included, or other incorrect combinations of signing fields — this is not a dedicated check; it is caught by the same signature-validity logic used elsewhere ("cannot both single- and multi-sign"), so it surfaces as `temBAD_SIGNATURE`, not a distinct malformed error.
+1. `Sponsor`, `SponsorFlags`, or `SponsorSignature` is included in a transaction that does not support sponsorship (see section [8.3.4](#834-transactions-that-cannot-be-sponsored)) (`temINVALID_FLAG`).
+1. Exactly one of `Sponsor` and `SponsorFlags` is included (they must either both be included, if the transaction is sponsored, or neither, if it is not) (`temINVALID_FLAG`).
+1. `SponsorSignature` is included but `Sponsor` or `SponsorFlags` is missing (`temMALFORMED`).
+1. `SponsorFlags` includes invalid flags — currently, the only two valid flags are `spfSponsorFee` and `spfSponsorReserve` (`temINVALID_FLAG`).
+1. `Sponsor` is equal to `tx.Account` (an account may not sponsor its own transaction) (`temMALFORMED`).
```
Tapanito: Please index these in an ascending order.

XLS-0068-sponsored-fees-and-reserves/README.md
```diff
@@ -536,13 +581,12 @@ If a `Sponsorship` object does not exist, the `tx.Fee` value is decremented from
...
+1. The sponsee's `SponsoredOwnerCount` field will be incremented by the number of objects that are sponsored as a part of the transaction.
+1. The `SponsoredOwnerCount`, `SponsoringOwnerCount`, and `SponsoringAccountCount` fields will be decremented when those objects/accounts are deleted.
+1. If a `Sponsorship` object exists, its `RemainingOwnerCount` is decremented by the same amount. **This budget is one-way: it is consumed when a sponsored object is created, but it is _not_ credited back when that object is later deleted, or when its sponsorship is transferred away via `SponsorshipTransfer`.** `RemainingOwnerCount` is therefore a cumulative allowance of sponsorships granted, not a live count of currently-sponsored objects; the sponsor's `SponsoringOwnerCount` (which _is_ decremented on deletion) is the live count. A sponsor who wants to extend the allowance must top it up explicitly with a positive `RemainingOwnerCountDelta` via `SponsorshipSet`.
```
Tapanito: Please reindex.

XLS-0068-sponsored-fees-and-reserves/README.md
```diff
@@ -828,8 +900,9 @@ Additional failure conditions when `tfSponsorCreatedAccount` is enabled:
 
 1. `tfNoRippleDirect`, `tfPartialPayment`, or `tfLimitQuality` are enabled (`temINVALID_FLAG`)
 1. `Amount` specifies a non-XRP currency (`temBAD_AMOUNT`)
+1. `SendMax` or `Paths` is included (`temINVALID`)
 1. `Destination` already exists (`tecNO_SPONSOR_PERMISSION`)
-1. `Account` does not have enough XRP to cover the account reserve requirement (`tecNO_DST_INSUF_XRP`)
+1. `Account` (the sponsor) does not have enough XRP to fund the payment (`tecUNFUNDED_PAYMENT`). Note: the usual `tecNO_DST_INSUF_XRP` check (that a direct-XRP payment to a new account must be at least the account reserve) is waived when this flag is set, not replaced by a new check — `tecUNFUNDED_PAYMENT` is simply the ordinary insufficient-funds failure that already applies to any direct XRP payment.
```
Tapanito: Please reindex.

XLS-0068-sponsored-fees-and-reserves/README.md
```diff
@@ -895,43 +964,101 @@ If the `AccountRoot` associated with the `tx.Account` has a `SponsoredOwnerCount
...
+- Where a `BatchSigners` entry is required per the rule above, the required set is matched exactly: every required sponsor must appear, no extras are permitted, and the array must be sorted and free of duplicates. One signer entry authorizes the sponsorship of every inner transaction that names that account as `Sponsor`. A sponsor that is the outer `Batch` account itself is not added to the required set — the outer account already signs the `Batch` — and in fact **must not** appear in `BatchSigners`.
+- Fee and reserve resolution for an inner transaction otherwise follows the standard rules in [sections 8.3.2](#832-fee-sponsorship-failures), [8.3.3](#833-reserve-sponsorship-failures), and [section 17](#17-feature-interactions).
+- Inner transactions **must not** have `SponsorFlags.spfSponsorFee` enabled, as the outer `Batch` transaction pays the fee for all inner transactions.
```
Tapanito, line 1017: Shouldn't this list be indexed numerically?

XLS-0068-sponsored-fees-and-reserves/README.md
```diff
@@ -442,7 +457,7 @@ There are two flag values that are supported:
 | `spfSponsorFee`     | `0x00000001` | Sponsoring (paying for) the fee of the transaction.                |
 | `spfSponsorReserve` | `0x00000002` | Sponsoring the reserve for any objects created in the transaction. |
 
-#### 8.1.2 `SponsorSignature`
+#### 8.1.2. `SponsorSignature`
 
 | Field Name      | Required? | JSON Type | Internal Type | Description                                                                                                                                           |
 | --------------- | --------- | --------- | ------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------- |
```
Tapanito: **Template — `SponsorSignature` is documented as a nested table, and the common fields table above omits `Default Value`.**

In `AMENDMENT_TEMPLATE.md` an inner object's shape belongs in a `Field Details` subsection (the 3.1.1 style) rather than a second table inline, and the transaction field table includes a `Default Value` column that the table at line 443 does not have.

The content is right, only the presentation differs: `sfSponsorSignature` is registered in `InnerObjectFormats.cpp` with `SigningPubKey`, `TxnSignature` and `Signers` all optional, and it is `kNotSigning` in `sfields.macro`, matching section 8.1.2.1.

XLS-0068-sponsored-fees-and-reserves/README.md
```diff
@@ -581,22 +627,27 @@ This transaction uses the standard transaction fee (currently 10 drops, subject
 
 ### 9.4. Failure Conditions
```
Tapanito: **Blocking (template) — Failure Conditions are not split into the two required subsections.**

`AMENDMENT_TEMPLATE.md` section 3.4 requires `3.4.1 Data Verification` for `tem` codes (checkable without ledger state) and `3.4.2 Protocol-Level Failures` for `tec` / `ter` / `tef` / `tel`. This section uses one flat numbered list with both classes interleaved — item 6 (`temBAD_AMOUNT`) and item 8 (`tecNO_DST`) sit side by side.

The same applies to every other failure-conditions list in the spec: section 8.3 (line 495), 10.5 (line 810), 11.3 (line 893), 12.2 (line 937) and 13.4 (line 993).

The codes themselves check out against develop — `temREDUNDANT`, `tecPSEUDO_ACCOUNT`, `tecLIMIT_EXCEEDED`, `tecUNFUNDED`, `tecNO_ENTRY` and `tecNO_DST` are all present in `SponsorshipSet.cpp`.

XLS-0068-sponsored-fees-and-reserves/README.md
```diff
@@ -744,44 +809,51 @@ This transaction uses the standard transaction fee (currently 10 drops, subject
...
 
 Additional failure conditions specific to `SponsorshipTransfer`:
 
 1. Exactly one flag must be set (`temINVALID_FLAG`)
-2. The set of fields included in the transaction does not match the expected set for the specified flag (`temMALFORMED`)
-3. `ObjectID` is specified but does not exist on the ledger (`tecNO_TARGET`)
+2. The set of fields included in the transaction does not match the expected set for the specified flag. The exact error code depends on which field is misused: `Sponsor` present without `spfSponsorReserve` on the Create/Reassign path, or `SponsorFlags` present on the End path (`temINVALID_FLAG`); `Sponsee` present on the Create/Reassign path, `Sponsor` present on the End path, or `Sponsee` equal to `tx.Account` on the End path (`temMALFORMED`)
+3. `ObjectID` is specified but does not exist on the ledger (`tecNO_ENTRY`)
 4. `ObjectID` is specified but does not have a `Sponsor` field (object is not sponsored) (`tecNO_PERMISSION`)
 5. `ObjectID` is not specified and the `tx.Account` does not have a `Sponsor` field (account is not sponsored) (`tecNO_PERMISSION`)
```
Tapanito: **Blocking (implementation) — these two items contradict the Create path.**

Both items say a target without a `Sponsor` field fails with `tecNO_PERMISSION`, and they are stated as general conditions ahead of the per-flag lists in items 8–10.

On develop, `SponsorshipTransfer::preclaim` only requires an existing sponsor on the End and Reassign paths. For `tfSponsorshipCreate` the target must **not** already be sponsored:

```cpp
if (!newSponsorSle || isSponsored)
    return tecNO_PERMISSION;
```

As written, the general list makes the Create happy path a failure, and it contradicts section 10.3.2, which says the object "must **not** have a `Sponsor` field".

Either move these two under the End/Reassign sub-lists, or reword to "the target's sponsorship state must match the flag".

XLS-0068-sponsored-fees-and-reserves/README.md
```diff
@@ -895,43 +964,101 @@ If the `AccountRoot` associated with the `tx.Account` has a `SponsoredOwnerCount
...
+        "Sequence": 0,
+        "Fee": "0",
+        "SigningPubKey": ""
+      }
+    }
+  ],
+  "BatchSigners": [
+    {
+      "BatchSigner": {
+        "Account": "rN7n7otQDd6FczFgLdlqtyMVrn3HMfXpf",
```
Tapanito: **Blocking (implementation) — this example would be rejected with `temBAD_SIGNER`.**

The outer `Account`, the inner transaction's `Sponsor` (line 1025) and this `BatchSigners[0].Account` are all `rN7n7otQDd6FczFgLdlqtyMVrn3HMfXpf`.

On develop, `Batch::preflightSigValidated` rejects any signer equal to the outer account, and it also excludes a sponsor that is the outer account from `requiredSigners`. So this batch both contains a forbidden entry and needs no `BatchSigners` at all. Section 13.3 states the rule correctly; only the example contradicts it.

XLS-0068-sponsored-fees-and-reserves/README.md
```diff
@@ -999,6 +1136,21 @@ There are no additional failure conditions.
 }
 ```
 
+### 15.6. Related: `ledger_entry` Sponsorship Lookup
```
Tapanito: **Template — the `ledger_entry` change should be its own `RPC:` section.**

It is documented here as a note under `account_objects`, but it is a real API change: `LedgerEntry.cpp` adds `parseSponsorship`, which reads `jss::sponsor` and `jss::sponsee` and resolves to `keylet::sponsorship`.

The template's `RPC:` section (Request Fields, Response Fields, Failure Conditions, Example Request, Example Response) is where a reviewer will look for it, and the error codes it can return — `malformedSponsor` and `malformedSponsee` — are currently undocumented.

Tapanito (changes requested):
Review of XLS-68 in two passes: template conformity against `templates/XLS_TEMPLATE.md` and `templates/AMENDMENT_TEMPLATE.md`, and implementation correctness against rippled `develop` (Sponsor landed via [XRPLF/rippled#7350](https://github.com/XRPLF/rippled/pull/7350)).

The spec tracks the merged implementation closely. Amendment name and vote behavior, `ltSPONSORSHIP = 0x0090` and the `'>'` keylet namespace, the `Sponsorship` field set, the ledger and transaction flag values, the `SponsorshipSet` delta-and-clamp semantics, the reserve allow-list, the fee-payer precedence rules, the one-way `RemainingOwnerCount`, the invariants in section 19, and the `account_objects` / `ledger_entry` changes all match what is on develop. The findings below are the places where the text and the code disagree, plus the template deviations.

Most of the substance is in the inline comments. Five findings landed on lines outside the diff hunks, so they are here instead.

---

**Blocking — section structure does not follow `XLS_TEMPLATE.md` (line 30).**

The template fixes the top-level sequence: 1 Abstract, 2 Motivation, 3 Specification (with the amendment template nested underneath), 4 Rationale, 5 Backwards Compatibility, 6 Test Plan, 7 Reference Implementation, 8 Security Considerations, then the optional appendix. This document has no `## 3. Specification` at all — Overview, the ledger entry, transaction and RPC sections, Feature Interactions and Example Flows all sit at H2 — and Security Considerations (section 18) comes before Rationale (section 21). `XLS-0065-single-asset-vault/README.md` is the shape to match.

This one is structural, so it is worth deciding deliberately: restructure XLS-68 now, or land it as-is and take the template debt in a follow-up.

---

**Should-fix — two `SponsorshipSet` flag descriptions say "fees" where they mean "reserves" (lines 618–619).**

`tfSponsorshipSetRequireSignForReserve` and `tfSponsorshipClearRequireSignForReserve` are both described as governing "sponsoring fees". The flag values themselves (`0x00040000`, `0x00080000`) match `TxFlags.h`.

---

**Should-fix — `RippleState` example JSON has sponsors but not the matching reserve flags (line 427).**

The example sets both `HighSponsor` and `LowSponsor` with `"Flags": 262144`, which is `0x00040000` (`lsfLowAuth`). Section 7.2 requires `lsfLowReserve` for `LowSponsor` and `lsfHighReserve` for `HighSponsor`, i.e. `0x00010000` and `0x00020000`, so `196608` for both.

---

**Should-fix — broken cross-reference in section 10.1.2 (line 746).**

The text points at "section 6.1.1" / `#611-sponsor`, which is `AccountRoot.Sponsor`. The transaction common field `Sponsor` is section 8.1 (`#81-fields`).

---

**Should-fix — FAQ A.16 contradicts section 7.1 (line 1839).**

A.16 says the trust-line case "is still being explored" and floats a `Sponsor2` field. Section 7.1 already specifies `HighSponsor` / `LowSponsor`, and that is what `ledger_entries.macro` implements for `ltRIPPLE_STATE` on develop.

---

One note on the field tables, since it spans several sections and only one is anchorable: the ledger entry tables (4.1, 5.2, 6.1, 7.1) add a `JSON Type` column and put `Default Value` before the types, while the transaction tables (8.1, 9.1, 10.1) drop `Default Value` entirely. The template is `Field Name | Constant | Required | Internal Type | Default Value | Description` throughout.

Tapanito (approved):

PR #598, opened 2026-08-07
XLS-99: Confidential MPT Key Rotation

yinyiqian1:
### Type of Change

- [X] New XLS Draft


mvadari (dismissed):
Please follow the format for drafts expected in the CONTRIBUTING document and the templates (namely, the file naming)

XLS-draft-confidential-mpt-key-rotation/README.md
```diff
@@ -0,0 +1,1046 @@
...
+| :---------------- | :-------- | :-------- | :------------ | :------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| `IssuerKeyEpoch`  | No        | `number`  | `UINT32`      | Monotonically increasing counter incremented on each issuer ElGamal key rotation. Not stored when at default value 0. Validators treat an absent field as epoch 0.  |
+| `AuditorKeyEpoch` | No        | `number`  | `UINT32`      | Monotonically increasing counter incremented on each auditor ElGamal key rotation. Not stored when at default value 0. Validators treat an absent field as epoch 0. |
+
+**Note**: To accommodate existing `MPTokenIssuance` ledger objects that lack epoch fields even when keys are registered, the epoch value should remain absent after initial registration. It is set to 1 only when rotating a key for the first time successfully, and then increments with each subsequent rotation.
+
```
Tapanito: Please update the table to follow this format exactly: https://github.com/XRPLF/XRPL-Standards/blob/master/templates/AMENDMENT_TEMPLATE.md#22-fields

The table is missing:
* `Constant` column
* The `?` (question mark) must be dropped from `Required` column
* `Default Value` column

XLS-draft-confidential-mpt-key-rotation/README.md
```diff
@@ -0,0 +1,1046 @@
...
+}
+```
+
+## 7. Ledger Entry: `MPToken`
+
+The existing `MPToken` ledger object is extended with three new fields. All other fields, flags, ownership, reserves, and the object identifier are unchanged from XLS-0033.
+
+### 7.1. Fields
+
+| Field Name              | Required? | JSON Type | Internal Type | Description                                                                                                                                                                                                                                                                                                                                                                                 |
```
Tapanito: Please update the columns

XLS-draft-confidential-mpt-key-rotation/README.md
```diff
@@ -0,0 +1,1046 @@
...
+3.  Pre-`ConfidentialMPTKeyRotation` amendment, an issuer cannot register an auditor key in `MPTokenIssuanceSet` unless the issuer key is being registered in the same transaction. This means
+    the issuer can either register the issuer key alone or register both keys together initially. With the `ConfidentialMPTKeyRotation` amendment, the issuer can now register an auditor key at any time after the issuer key is already registered or rotated—allowing them to opt in whenever they want. The only constraint is that an auditor key cannot be registered before an issuer key.
+
+The guard against adding `IssuerEncryptionKey` when `lsfMPTCanHoldConfidentialBalance` is not enabled remains unchanged - confidential transfers must be enabled before keys can be set or rotated.
+All existing `MPTokenIssuanceSet` behavior (lock/unlock, `DomainID`) is completely unaffected.
+
+### 8.1. Fields
+
+| Field Name             | Required? | JSON Type | Internal Type | Description                                                                                                                                                                                                   |
```
Tapanito: This table is missing the `Default Value` column.

XLS-draft-confidential-mpt-key-rotation/README.md
```diff
@@ -0,0 +1,1046 @@
...
+
+5. **Holder Issuer-Mirror Self-Migration**: Re-encrypts issuer mirror.
+6. **Holder Auditor-Mirror Self-Migration**: Re-encrypts the auditor mirror, or sets it for the first time if late-registered.
+7. **Simultaneous Holder Self-Migration**: Updates both the issuer and auditor encrypted balances in a single transaction (used when both keys have rotated).
+
+### 9.2. Fields
+
+Whether it is issuer mode or holder mode is determined by `Holder` field's presence. If `Holder` is present, it is issuer mode. If `Holder` is absent, it is holder self-migration mode.
+
+| Field Name                    | Required? | JSON Type | Internal Type | Description                                                                                                                                                                                                                         |
```
Tapanito: Missing `Default Value` column

XLS-draft-confidential-mpt-key-rotation/README.md
```diff
@@ -0,0 +1,1046 @@
+<pre>
+  xls: ??
+  title: ElGamal Key Rotation for Confidential MPTs
+  description: Defines ElGamal key rotation for issuer, auditor, and holder roles in the Confidential MPT protocol, with key loss recovery mechanisms.
+  author: Aanchal Malhotra <amalhotra@ripple.com> Yinyi Qian <yqian@ripple.com>
+  category: Amendment
+  status: Draft
+  requires: XLS-0096
```
Tapanito: The spec also requires XLS-33

Tapanito (changes requested):

Tapanito:
## Complexity concern: mode explosion, especially in `ConfidentialMPTMirrorUpdate`

Overall the three-role rotation design (issuer/auditor two-phase, holder self-contained, holder-loss two-step recovery) is a reasonable shape for the problem. But a few places look accretive rather than designed top-down, and I'd like the authors to consider tightening before this goes further:

**1. `ConfidentialMPTMirrorUpdate` has too many modes for one transaction type.**
Mode is selected by `Holder` presence (issuer vs. holder-self-migration), further split by which of `IssuerEncryptedAmount`/`AuditorEncryptedAmount` are present (issuer-only / auditor-only / both) — 6-7 distinct behaviors, each discharged by its own sigma protocol (Section 9.7). Two of those relations ("holder both-mirrors" and "auditor-only") are explicitly "not yet covered" by the companion proof spec. That's a lot of surface for a transaction that's meant to do one conceptual thing ("re-encrypt a stale mirror").

Compare this to `ConfidentialMPTHolderKeyUpdate`, which deliberately uses **explicit flags** rather than field-presence overloading for mode selection, and the rationale given (18.5) is that flags "make validator logic unambiguous and eliminate edge cases with partial field sets." That rationale applies with equal force to `ConfidentialMPTMirrorUpdate`, which doesn't follow it. Suggest either:
- splitting issuer-mode and holder-self-migration-mode into two transaction types (mirroring the flag-based approach), or
- collapsing issuer-only/auditor-only/both into a single relation by always requiring both encrypted-amount fields, with a no-op/null delta for the mirror not being rotated.

**2. The same staleness invariant is enforced in at least three separate places** — validator preclaim (4 existing transactions × 2 conditions each, Section 14.1), wallet-side pre-submission checks (14.3), and off-chain regulator verification (14.5) — with no single source of truth beyond "compare two epoch fields." This isn't wrong, but it's a sign the epoch model is being threaded through the whole system rather than contained. Worth double-checking all three call sites stay in sync as the spec evolves.

**3. Historical key retention is a correctness requirement with no protocol enforcement.** Per 15.9, if the issuer destroys a pre-rotation secret key before all holders at that epoch have migrated, those holders lose the actively-migrated path permanently and must fall back to self-migration — which requires the holder to have `sk_H` and to have already merged their inbox. There's no on-chain signal that would stop an issuer from destroying a key too early, and no enforcement mechanism suggested beyond "issuers should retain historical secret keys." For a protocol otherwise built on formal ZK guarantees, this correctness property resting entirely on issuer operational discipline stands out.

**4. Issuer-key-loss recovery has a real, acknowledged gap for inactive holders** (13.5) with no on-chain fallback — permanent suspension of clawback authority is listed as an accepted outcome. Given how much of the rest of the spec is built specifically to close comparable gaps (e.g. the entire holder-key-loss recovery flow exists to solve exactly this class of problem for the *holder* side), it's worth asking whether the same asymmetry is acceptable for the *issuer* side, or whether it deserves the same design attention.

**5. Permission delegation for `ConfidentialMPTHolderKeyUpdate` recovery mode is left as "TBD" (17.2/17.3)** in a spec that's otherwise fully specified down to proof byte-sizes. Recommend resolving this before merge — it's a security-relevant decision (whether a custody provider can trigger recovery without holder consent), not an implementation detail.

Given the number of interacting moving parts (5 new ledger fields, 8 new invariants, 4 new/modified transactions, 6+ proof relations), I'd suggest the authors provide a state-machine diagram per role (issuer, auditor, holder) showing valid epoch/mirror-epoch transitions — it would make it much easier to verify there's no missed corner case, and would likely surface further opportunities to collapse modes.

XLS-draft-confidential-mpt-key-rotation/README.md
yinyiqian1: https://github.com/XRPLF/XRPL-Standards/pull/598/changes/cd399ea70aab6fa2163d1c11f4cb787bd8eba001

XLS-draft-confidential-mpt-key-rotation/README.md
yinyiqian1: https://github.com/XRPLF/XRPL-Standards/pull/598/changes/cd399ea70aab6fa2163d1c11f4cb787bd8eba001

XLS-draft-confidential-mpt-key-rotation/README.md
yinyiqian1: https://github.com/XRPLF/XRPL-Standards/pull/598/changes/cd399ea70aab6fa2163d1c11f4cb787bd8eba001

XLS-draft-confidential-mpt-key-rotation/README.md
yinyiqian1: https://github.com/XRPLF/XRPL-Standards/pull/598/changes/cd399ea70aab6fa2163d1c11f4cb787bd8eba001

XLS-draft-confidential-mpt-key-rotation/README.md
yinyiqian1: https://github.com/XRPLF/XRPL-Standards/pull/598/changes/cd399ea70aab6fa2163d1c11f4cb787bd8eba001

yinyiqian1:
> ## Complexity concern: mode explosion, especially in `ConfidentialMPTMirrorUpdate`
> Overall the three-role rotation design (issuer/auditor two-phase, holder self-contained, holder-loss two-step recovery) is a reasonable shape for the problem. But a few places look accretive rather than designed top-down, and I'd like the authors to consider tightening before this goes further:
> 
> **1. `ConfidentialMPTMirrorUpdate` has too many modes for one transaction type.** Mode is selected by `Holder` presence (issuer vs. holder-self-migration), further split by which of `IssuerEncryptedAmount`/`AuditorEncryptedAmount` are present (issuer-only / auditor-only / both) — 6-7 distinct behaviors, each discharged by its own sigma protocol (Section 9.7). Two of those relations ("holder both-mirrors" and "auditor-only") are explicitly "not yet covered" by the companion proof spec. That's a lot of surface for a transaction that's meant to do one conceptual thing ("re-encrypt a stale mirror").
> 
> Compare this to `ConfidentialMPTHolderKeyUpdate`, which deliberately uses **explicit flags** rather than field-presence overloading for mode selection, and the rationale given (18.5) is that flags "make validator logic unambiguous and eliminate edge cases with partial field sets." That rationale applies with equal force to `ConfidentialMPTMirrorUpdate`, which doesn't follow it. Suggest either:
> 
> * splitting issuer-mode and holder-self-migration-mode into two transaction types (mirroring the flag-based approach), or
> * collapsing issuer-only/auditor-only/both into a single relation by always requiring both encrypted-amount fields, with a no-op/null delta for the mirror not being rotated.
> 
> **2. The same staleness invariant is enforced in at least three separate places** — validator preclaim (4 existing transactions × 2 conditions each, Section 14.1), wallet-side pre-submission checks (14.3), and off-chain regulator verification (14.5) — with no single source of truth beyond "compare two epoch fields." This isn't wrong, but it's a sign the epoch model is being threaded through the whole system rather than contained. Worth double-checking all three call sites stay in sync as the spec evolves.
> 
> **3. Historical key retention is a correctness requirement with no protocol enforcement.** Per 15.9, if the issuer destroys a pre-rotation secret key before all holders at that epoch have migrated, those holders lose the actively-migrated path permanently and must fall back to self-migration — which requires the holder to have `sk_H` and to have already merged their inbox. There's no on-chain signal that would stop an issuer from destroying a key too early, and no enforcement mechanism suggested beyond "issuers should retain historical secret keys." For a protocol otherwise built on formal ZK guarantees, this correctness property resting entirely on issuer operational discipline stands out.
> 
> **4. Issuer-key-loss recovery has a real, acknowledged gap for inactive holders** (13.5) with no on-chain fallback — permanent suspension of clawback authority is listed as an accepted outcome. Given how much of the rest of the spec is built specifically to close comparable gaps (e.g. the entire holder-key-loss recovery flow exists to solve exactly this class of problem for the _holder_ side), it's worth asking whether the same asymmetry is acceptable for the _issuer_ side, or whether it deserves the same design attention.
> 
> **5. Permission delegation for `ConfidentialMPTHolderKeyUpdate` recovery mode is left as "TBD" (17.2/17.3)** in a spec that's otherwise fully specified down to proof byte-sizes. Recommend resolving this before merge — it's a security-relevant decision (whether a custody provider can trigger recovery without holder consent), not an implementation detail.
> 
> Given the number of interacting moving parts (5 new ledger fields, 8 new invariants, 4 new/modified transactions, 6+ proof relations), I'd suggest the authors provide a state-machine diagram per role (issuer, auditor, holder) showing valid epoch/mirror-epoch transitions — it would make it much easier to verify there's no missed corner case, and would likely surface further opportunities to collapse modes.


1. Prefer not to apply.
`ConfidentialMPTHolderKeyUpdate` requires two explicit flags for unambiguous anchoring because the holder is always the transaction submitter. In contrast, `ConfidentialMPTMirrorUpdate` distinguishes between holder and issuer modes directly via the `Holder` field, since the submitter is the issuer in issuer mode and the holder in holder mode. Splitting issuer mode and holder self-migration into separate transaction types would not eliminate most of the complexity. Both transaction types would still need issuer-only, auditor-only, and both-mirror variants, so the six proof relations would become two groups of three rather than being reduced.

2. Simplified the mirror staleness instructions. Section 5.3.1 now defines mirror currentness as the single source of truth, while Section 5.3.2 lists the transactions that require current mirrors. Removed the redundant validator, wallet, and regulator descriptions. https://github.com/XRPLF/XRPL-Standards/pull/598/changes/8fcf1b9550e909a8d14d94c3505679f62f13b868

3. Clarified Section 15.9 that historical issuer-key retention is an operational migration choice, not a correctness requirement. If the issuer makes sure to migrate all the holders, it does not need the historical keys. https://github.com/XRPLF/XRPL-Standards/pull/598/changes/da53a1d85ce26e5d23b8a789cb451e2f786976f9

4. Clarified Section 13.5 to make the counterparty dependency explicit. https://github.com/XRPLF/XRPL-Standards/pull/598/changes/8fcf1b9550e909a8d14d94c3505679f62f13b868

5. Updated the permission delegation decision: https://github.com/XRPLF/XRPL-Standards/pull/598/changes/3cc968c81c6e498a5c1e0a625464b79145937f21

Added an informative state-transition summary in Section 5.4. Rather than using three role-specific diagrams, which would duplicate the shared issuer/auditor mirror lifecycle, added two transition tables: https://github.com/XRPLF/XRPL-Standards/pull/598/changes/fd7429b2d6ff32dea85e07268b853a6449a507a7
- Mirror lifecycle: missing, current, and stale mirror transitions.
- Holder key and recovery lifecycle: no-recovery-pending and recovery-pending transitions.

PeterChen13579 (approved):
Looks good for a draft. We'll continue to update this spec as we update the code as we go 👍

XLS-draft-confidential-mpt-key-rotation/README.md
```diff
@@ -0,0 +1,1046 @@
+<pre>
+  xls: ??
+  title: ElGamal Key Rotation for Confidential MPTs
+  description: Defines ElGamal key rotation for issuer, auditor, and holder roles in the Confidential MPT protocol, with key loss recovery mechanisms.
+  author: Aanchal Malhotra <amalhotra@ripple.com> Yinyi Qian <yqian@ripple.com>
```
Tapanito: The XLS preamble expects `author` to be a comma-separated list. Right now the two authors run together, so a parser reading this field sees one entry.

```suggestion
  author: Aanchal Malhotra <amalhotra@ripple.com>, Yinyi Qian <yqian@ripple.com>
```

XLS-draft-confidential-mpt-key-rotation/README.md
```diff
@@ -0,0 +1,1046 @@
...
+  - Recovery mode (`tfHolderKeyRecovery`): holder has lost `sk_H`; registers new key as `RecoveryKey` for issuer-completed recovery.
+  - Cancel mode (`tfCancelRecovery`): holder clears a pending `RecoveryKey` authorization.
+- `ConfidentialMPTRecoverBalance`: Completes holder key loss recovery by re-encrypting balances under the authorized new key. Submitted by the issuer.
+
+### 4.3. Modified Ledger Entries
+
+- `MPTokenIssuance`: Two new fields - `IssuerKeyEpoch`, `AuditorKeyEpoch`.
+- `MPToken`: Three new fields - `IssuerKeyMirrorEpoch`, `AuditorKeyMirrorEpoch`, `RecoveryKey`.
+
+## 5. Specification
```
Tapanito: `XLS_TEMPLATE.md` nests the amendment scaffold **inside** this Specification section. As written, Specification is a narrative overview and the ledger entries and transactions become top-level `## 6`–`## 11` peers of Abstract.

Could the ledger entries, transactions, permissions, and any RPC deltas move under this heading, so the document reads as Abstract → Motivation → Specification (scaffold) → Rationale → Backwards Compatibility → Test Plan → Security Considerations? The key-rotation model, re-encryption strategy, and state transition tables are worth keeping — they just fit better as an overview subsection or an appendix than as a sibling of the normative sections.

XLS-draft-confidential-mpt-key-rotation/README.md
```diff
@@ -0,0 +1,1046 @@
...
+
+## 4. Scope
+
+### 4.1. Modified Transaction Types
+
+- `MPTokenIssuanceSet`: Extended to allow replacement of `IssuerEncryptionKey` and `AuditorEncryptionKey` when already present, enabling issuer and auditor key rotation.
+- `ConfidentialMPTSend`: Rejected when the sender's or the destination's issuer or auditor mirror is stale.
+- `ConfidentialMPTConvert`: Rejected when an already-initialized holder's issuer or auditor mirror is stale.
+- `ConfidentialMPTConvertBack`: Rejected when the holder's issuer or auditor mirror is stale.
+- `ConfidentialMPTClawback`: Rejected when the holder's issuer mirror is stale. See Section 12.1.
```
Tapanito: These four transactions gain normative behavior from this amendment: staleness rejection with `tecNO_PERMISSION`, and for clawback, mirror ciphertext resets plus epoch rewrites. Right now that is spread across this bullet list, section 5.3.2, and section 12.1.

The amendment template asks for the same subsections when you change an existing transaction type, with Transaction Fee optional. Please give each of the four its own `## Transaction: <name>` section carrying at least Failure Conditions and State Changes, the way `MPTokenIssuanceSet` already does in section 8. Implementers reading the spec per transaction type will otherwise miss the clawback epoch resets entirely.

XLS-draft-confidential-mpt-key-rotation/README.md
```diff
@@ -0,0 +1,1046 @@
...
+| :---------------- | :------- | :------- | :------------ | :------------ | :-------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `IssuerKeyEpoch`  | No       | No       | `UINT32`      | `0`           | Monotonically increasing counter incremented on each issuer ElGamal key rotation. Not stored when at its default value; validators treat an absent field as epoch 0.  |
+| `AuditorKeyEpoch` | No       | No       | `UINT32`      | `0`           | Monotonically increasing counter incremented on each auditor ElGamal key rotation. Not stored when at its default value; validators treat an absent field as epoch 0. |
+
+**Note**: To accommodate existing `MPTokenIssuance` ledger objects that lack epoch fields even when keys are registered, the epoch value should remain absent after initial registration. It is set to 1 only when rotating a key for the first time successfully, and then increments with each subsequent rotation.
+
+### 6.2. Invariants
```
Tapanito: The template asks for invariants as before/after statements using `<object>` for the before state and `<object>'` for the after state. I1–I4 are static well-formedness checks, which are useful but do not capture the transition this amendment actually introduces.

Worth adding alongside them, in template form:

- On a successful issuer key rotation: `<MPTokenIssuance>'.IssuerKeyEpoch == <MPTokenIssuance>.IssuerKeyEpoch + 1` (absent is treated as 0, so the first rotation yields 1).
- `<MPTokenIssuance>'.IssuerKeyEpoch >= <MPTokenIssuance>.IssuerKeyEpoch` — epochs never decrease.

The same applies to the auditor pair, and to I5/I6 in section 7.3.

XLS-draft-confidential-mpt-key-rotation/README.md
```diff
@@ -0,0 +1,1046 @@
...
+| `AuditorKeyMirrorEpoch` | No       | No       | `UINT32`      | `0`           | The `AuditorKeyEpoch` at which this holder's auditor mirror was last re-encrypted. The mirror is stale when this value is less than the current `AuditorKeyEpoch`. Not stored when at its default value.                                                                                                                                                                                    |
+| `RecoveryKey`           | No       | No       | `BLOB`        | N/A           | A 33-byte compressed ElGamal public key authorized for key loss recovery. Set by `ConfidentialMPTHolderKeyUpdate` in recovery mode (`tfHolderKeyRecovery`). Cleared by exactly two paths: (1) `ConfidentialMPTRecoverBalance` when the issuer completes recovery; (2) `ConfidentialMPTHolderKeyUpdate` with `tfCancelRecovery` when the holder explicitly cancels. Has no automatic expiry. |
```
Tapanito: `RecoveryKey` carries a lot of behavior in one table cell: who may set it, the two clearing paths, the absence of any expiry, the 33-byte compressed point requirement, and the constraint that it must differ from the current `HolderEncryptionKey`.

The template has a Field Details subsection (2.2.1) for exactly this. Please promote it, and do the same for `IssuerKeyMirrorEpoch` and `AuditorKeyMirrorEpoch` — the rule that an absent field means epoch 0, that initial registration leaves the epoch absent, and that the first rotation creates the field with value 1 is currently split between this table and the note under section 6.1.

XLS-draft-confidential-mpt-key-rotation/README.md
```diff
@@ -0,0 +1,1046 @@
...
+  "Flags": 0,
+  "MPTokenIssuanceID": "000000012A9F1D3C...",
+  "Account": "rHolderAccountAddress",
+  "HolderEncryptionKey": "02c1d2e3f4a5b6...",
+  "ConfidentialBalanceSpending": "02d1e2f3a4b5c6...",
+  "ConfidentialBalanceInbox": "02e1f2a3b4c5d6...",
+  "ConfidentialBalanceVersion": 3,
+  "IssuerEncryptedBalance": "02f1a2b3c4d5e6...",
+  "IssuerKeyMirrorEpoch": 1,
+  "RecoveryKey": "02g1h2i3j4k5l6...",
```
Tapanito: `02g1h2i3j4k5l6...` is not valid hex — `g`, `h`, `i`, `j`, `k`, and `l` are outside the hex alphabet. Since this field is specified as a 33-byte compressed secp256k1 point, the example should at least be decodable as hex.

```suggestion
  "RecoveryKey": "02a9b8c7d6e5f4...",
```

XLS-draft-confidential-mpt-key-rotation/README.md
```diff
@@ -0,0 +1,1046 @@
...
+| `AuditorEncryptedAmount`      | Conditional | `string`  | `BLOB`        | N/A                           | A 66-byte ElGamal ciphertext encrypting the holder's balance under the new auditor key. Reuses `sfAuditorEncryptedAmount` from XLS-0096. Present to migrate holder's auditor mirror. At least one of this field and `IssuerEncryptedAmount` must be present. |
+| `PreviousIssuerEncryptionKey` | Conditional | `string`  | `BLOB`        | N/A                           | A 33-byte compressed ElGamal public key: the issuer key the holder's existing `sfIssuerEncryptedBalance` was encrypted under. **Required** when `Holder` and `IssuerEncryptedAmount` are both present, and **forbidden** otherwise.                          |
+| `ZKProof`                     | Yes         | `string`  | `BLOB`        | N/A                           | A single compact Chaum-Pedersen equality proof proving the new ciphertext(s) encrypt the same value as the on-ledger mirror(s). When both fields are present, the proof covers both statements under one Fiat-Shamir challenge.                              |
+
+### 9.3. Transaction Fee
```
Tapanito: Template section 3.3 wants an explicit **Fee Structure: Standard** or **Custom**, followed by the reason. "10x base fee, consistent with XLS-0096 confidential transactions" states the value but not the structure, and it leaves open whether this is 10x the current base fee at the time of writing or 10x whatever the base fee is under fee voting.

Suggested shape:

> **Fee Structure:** Custom
>
> This transaction requires 10x the base fee because it carries a 128-byte zero-knowledge proof requiring elliptic curve verification, consistent with the XLS-0096 confidential transactions. The multiplier applies to the current base fee and therefore tracks Fee Voting changes.

Same applies to sections 10.4 and 11.2.

XLS-0099-confidential-mpt-key-rotation/README.md
```diff
@@ -0,0 +1,1046 @@
...
+
+### 9.6. Example JSON
+
+Which of `IssuerEncryptedAmount` and `AuditorEncryptedAmount` a transaction
+carries depends on the use case, at least one must be present, and both may be.
+`PreviousIssuerEncryptionKey` follows from that: it is required whenever the
+transaction is in issuer mode and carries `IssuerEncryptedAmount`, and must be
+absent in every other case.
+
+Issuer mode:
```
Tapanito, line 540: Both example transactions in this section omit `Fee` and `Sequence`. The template's example (3.7) includes them, and the `MPTokenIssuanceSet` example in section 8.4 already does — so this is inconsistent within the document as well. The `ConfidentialMPTRecoverBalance` example around line 706 has the same gap.

XLS-draft-confidential-mpt-key-rotation/README.md
```diff
@@ -0,0 +1,1046 @@
...
+
+1. `RecoveryKey` on `MPToken` ← `HolderEncryptionKey` (new value)
+2. All other fields unchanged
+
+**Cancel mode**:
+
+1. `RecoveryKey` on `MPToken` ← cleared (field removed)
+2. All other fields unchanged - `HolderEncryptionKey`, `ConfidentialBalanceSpending`, `ConfidentialBalanceInbox`, `ConfidentialBalanceVersion` are not modified
+
+### 10.7. Example JSON
```
Tapanito: This section covers rotation and recovery modes but not cancel mode, which is the one with the most surprising shape — no `HolderEncryptionKey`, no `ZKProof`, no balance ciphertexts, just `Flags: 4`. Given that failure condition 10.5.1.7 rejects the transaction if any of those fields are present, an example is the clearest way to convey it:

```json
{
  "TransactionType": "ConfidentialMPTHolderKeyUpdate",
  "Account": "rHolderAccountAddress",
  "MPTokenIssuanceID": "000000012A9F1D3C...",
  "Flags": 4,
  "Fee": "120",
  "Sequence": 48
}
```

XLS-draft-confidential-mpt-key-rotation/README.md
```diff
@@ -0,0 +1,1046 @@
...
+  "TransactionType": "ConfidentialMPTRecoverBalance",
+  "Account": "rIssuerAccountAddress",
+  "MPTokenIssuanceID": "000000012A9F1D3C...",
+  "Holder": "rHolderAccountAddress",
+  "ConfidentialBalanceSpending": "02a1b2c3d4e5f6...",
+  "ZKProof": "03f1e2d3c4b5a6..."
+}
+```
+
+## 12. Clawback and Freeze Interactions with Key Rotation
```
Tapanito: The clawback rules here are transaction semantics: a precondition (issuer mirror must be current, `tecNO_PERMISSION` otherwise) and a set of state changes (mirror ciphertexts reset to canonical encrypted zero, mirror epochs set to the current key epochs).

Under the template those belong in the Failure Conditions and State Changes subsections of a `## Transaction: ConfidentialMPTClawback` section. Keeping them in a standalone section 12 means an implementer working through the clawback transaction has no pointer to the epoch resets.

XLS-draft-confidential-mpt-key-rotation/README.md
```diff
@@ -0,0 +1,1046 @@
...
+
+`ConfidentialMPTClawback` requires the holder's issuer mirror to be current, as
+specified in Section 5.3.2. On success, `IssuerEncryptedBalance` is reset to
+canonical encrypted zero under the current issuer key and
+`IssuerKeyMirrorEpoch` is set to `IssuerKeyEpoch`. If an auditor key is
+configured, `AuditorEncryptedBalance` is reset under the current auditor key
+and `AuditorKeyMirrorEpoch` is set to `AuditorKeyEpoch`. All other clawback
+semantics remain unchanged from XLS-0096.
+
+### 12.2. Freeze
```
Tapanito: The template has a Freeze/Lock subsection (2.8) on the ledger entry that holds the assets, which here is `MPToken`. The content of this section is right and the "locked holders may still rotate, because a locked account is still exposed to key compromise" rationale is a good call — it is just located away from the entry it describes.

Either move it under `Ledger Entry: MPToken` as Freeze/Lock, or add a short 2.8 there that states lock support is unchanged from XLS-0096 and links here.

XLS-draft-confidential-mpt-key-rotation/README.md
```diff
@@ -0,0 +1,1046 @@
...
+
+**If this check is missed**: There is no ambiguous state and no funds are at risk. The outcome depends on which transaction lands first:
+
+- **Rotation lands first**: The in-flight send's proof is bound to the old `ConfidentialBalanceVersion` and is rejected with `tecBAD_PROOF`. The holder's full balance is intact under pk_H'. The holder reconstructs the send proof under the new key and resubmits.
+- **Send lands first**: `ConfidentialBalanceVersion` bumps and `ConfidentialBalanceSpending` is debited normally. The rotation proof - constructed against the old version - is rejected with `tecBAD_PROOF`. The holder reconstructs the rotation proof against the updated state and resubmits.
+
+In both cases the result is a cleanly rejected transaction requiring resubmission - not a partial or ambiguous state. Clawback is unaffected in either scenario: `ConfidentialMPTClawback` uses `IssuerEncryptedBalance` (the issuer mirror), not `ConfidentialBalanceSpending` or `ConfidentialBalanceVersion`. The operational note is a UX concern, not a safety concern.
+
+### 14.3. Issuer Recovery Request Detection
```
Tapanito: This section, and section 5.3.5, both build on `mpt_holders`, and the Option 2 text says the per-holder lookups are needed "unless `mpt_holders` is extended to return `RecoveryKey`." That leaves the reader unsure whether this amendment changes an API or not.

Two clean resolutions:

1. No API change. Drop the extension clause, and move this operational guidance to the appendix so it is clearly non-normative.
2. Extend the API. Add a `## RPC: mpt_holders` section with Request Fields, Response Fields, Failure Conditions, and example request/response, per template section 5.

Either is fine; the current middle ground means the amendment's API surface is not pinned down.

XLS-draft-confidential-mpt-key-rotation/README.md
```diff
@@ -0,0 +1,1046 @@
...
+| `ConfidentialMPTMirrorUpdate`    | Issuer, one mirror    | Compact Chaum-Pedersen equality                                | 128 bytes |
+| `ConfidentialMPTMirrorUpdate`    | Issuer, both mirrors  | Compact Chaum-Pedersen (AND-composed)                          | 128 bytes |
+| `ConfidentialMPTMirrorUpdate`    | Holder self-migration | Cross-key equality proof                                       | 128 bytes |
+| `ConfidentialMPTHolderKeyUpdate` | Rotation              | Compact equality (spending) + new-key possession, AND-composed | 160 bytes |
+| `ConfidentialMPTHolderKeyUpdate` | Recovery              | Schnorr PoK                                                    | 64 bytes  |
+| `ConfidentialMPTRecoverBalance`  |                       | Compact Chaum-Pedersen equality                                | 128 bytes |
+
+All new transactions are charged 10x the base fee, consistent with XLS-0096.
+
+## 17. Permissions
```
Tapanito: The template models permissions as their own sections — `## Permission: [PermissionName]` with Permission Description, Transaction Types Affected, and Permission Value. This section is prose about how the new transactions interact with XLS-75, which covers the reasoning but not the template's contract.

The missing piece with real implementation impact is **Permission Value**: `ConfidentialMPTMirrorUpdate` and `ConfidentialMPTRecoverBalance` are declared delegable but have no numeric value assigned, and the `MPTokenIssuanceSet` restriction (a delegated submission must not carry `IssuerEncryptionKey` or `AuditorEncryptionKey`) is a new granular permission rule with no name or value.

Please either assign values in `Permission:` sections, or state explicitly that delegation reuses the transaction-type permission and give the integers. The non-delegable status of `ConfidentialMPTHolderKeyUpdate` is worth recording in the same structure so it is not read as an oversight.

Tapanito (changes requested):
## Review: amendment template compliance

This is a structural review against [`templates/AMENDMENT_TEMPLATE.md`](https://github.com/XRPLF/XRPL-Standards/blob/master/templates/AMENDMENT_TEMPLATE.md) as it is meant to be nested by [`templates/XLS_TEMPLATE.md`](https://github.com/XRPLF/XRPL-Standards/blob/master/templates/XLS_TEMPLATE.md). It is **not** a review of the cryptographic constructions.

The draft is already in the right family: `Ledger Entry:` / `Transaction:` headings, field tables with the template's columns, `tem*` vs `tec*` failure lists, state changes, and example JSON.

One caveat on tooling: `scripts/validate_xls_template.py` reports this file as valid, but that is not a green light. In my run every `xrpl.org` lookup returned HTTP 403, so the validator fell back to "assume the type already exists on mainnet" and skipped subsections that are required for **new** transactions.

`XLS_TEMPLATE.md` places the amendment scaffold **inside** `## 3. Specification`. This document instead follows the XLS-0096 outline, with a narrative Specification section and the ledger entries and transactions as top-level `## 6`–`## 11` peers of Abstract. The extra material (definitions, operational guidance, key-loss playbooks) is genuinely useful, but it should live in the outer XLS sections or an appendix rather than competing with the amendment heading contract.

### What already matches

- No new serialized types, so omitting the `SType` section is correct.
- `MPTokenIssuance` and `MPToken` are existing entries, so the optional Object Identifier, Ownership, Reserves, and RPC Name subsections are correctly omitted.
- Field table columns match the template for both ledger entries and transactions.
- All three new transactions have Fields, Failure Conditions (split into data verification and protocol level), State Changes, and Example JSON.
- `ConfidentialMPTHolderKeyUpdate` flag values are powers of two.
- Omitting the RPC section is correct, provided `mpt_holders` really is unchanged (see below).

### Requested changes

**1. Nest and retitle to the amendment scaffold.** Keep Abstract, Motivation, Definitions, and Scope as the XLS-1 introduction, then put ledger entries, transactions, permissions, and any RPC deltas under Specification using `## Ledger Entry: <code>Name</code>`, `## Transaction: <code>Name</code>`, `## Permission: <code>Name</code>`, and `## RPC: <code>method</code>`.

**2. Document the modified existing transactions as their own `Transaction:` sections.** The amendment adds normative behavior to `ConfidentialMPTSend`, `ConfidentialMPTConvert`, `ConfidentialMPTConvertBack`, and `ConfidentialMPTClawback` — staleness rejection with `tecNO_PERMISSION`, and mirror epoch rewrites on clawback. That currently lives in section 4.1, section 5.3.2, and section 12.1. The template asks for the same subsections when changing an existing type (Transaction Fee optional), so each of the four should get a delta section with at least Failure Conditions and State Changes, the way `MPTokenIssuanceSet` does.

**3. Permissions need the template shape.** Section 17 is prose about XLS-75. The template heading is `Permission: [PermissionName]` with Description, Transaction Types Affected, and Permission Value. Delegable new transaction types need numeric permission values, or a precise statement that they reuse the transaction-type permission and which integer that is.

**4. Transaction Fee wording.** For the new types, replace "10x base fee, consistent with XLS-0096" with the template's Fee Structure: Custom plus the reason, and note that it tracks the base fee under fee voting.

**5. Example JSON gaps.** Two examples omit `Fee` and `Sequence`, there is no cancel-mode example for `ConfidentialMPTHolderKeyUpdate`, and one `RecoveryKey` placeholder is not valid hex.

**6. Ledger entry polish.** Add Field Details subsections for `RecoveryKey` and the epoch fields, decide where Freeze/Lock lives, and state the invariants in the template's before/after form.

**7. Decide on the RPC.** Sections 5.3.5 and 14.3 both depend on `mpt_holders`, and 14.3 mentions extending it to return `RecoveryKey`. Either commit to no RPC change and drop the extension as a spec requirement, or add a full `RPC: mpt_holders` section.

**8. Outer XLS template.** Comma-separate the two authors, add Backwards Compatibility for relaxing the key-presence and COA > 0 guards on `MPTokenIssuanceSet`, and add a Test Plan — XLS-1 makes an implementation test plan mandatory for XLSes that affect `rippled`.

### Nits, non-blocking

- Transaction field tables list `Account` and `TransactionType`; the template treats common fields as assumed. Either drop them or keep them consistently across all four tables.
- Transaction naming is `ConfidentialMPT*` following XLS-0096 rather than the template's `<LedgerEntry><Verb>`. That is the right local convention — one sentence saying so heads off the nit.
- A one-line "no new metadata fields" would make that omission explicit.
- The companion proof specification is referenced several times but never linked.

### Suggested reshape

```text
## 5. Specification
  Ledger Entry: MPTokenIssuance
  Ledger Entry: MPToken
  Transaction: MPTokenIssuanceSet
  Transaction: ConfidentialMPTSend              # delta
  Transaction: ConfidentialMPTConvert           # delta
  Transaction: ConfidentialMPTConvertBack       # delta
  Transaction: ConfidentialMPTClawback          # delta
  Transaction: ConfidentialMPTMirrorUpdate
  Transaction: ConfidentialMPTHolderKeyUpdate
  Transaction: ConfidentialMPTRecoverBalance
  Permission: ... (values for delegable types, plus the issuance-set key restriction)
## 6. Rationale
## 7. Backwards Compatibility
## 8. Test Plan
## 9. Security Considerations
Appendix (operations, issuer key loss, FAQ, seed-derived keys)
```

Happy to re-review once the heading contract, the modified-transaction deltas, and the permission values are in place.

XLS-draft-confidential-mpt-key-rotation/README.md
yinyiqian1: https://github.com/XRPLF/XRPL-Standards/pull/598/changes/c30fed82daaa05ed89f9d4998de28b79cb0b68cd

XLS-draft-confidential-mpt-key-rotation/README.md
yinyiqian1: https://github.com/XRPLF/XRPL-Standards/pull/598/changes/13909cfd8f57ac95931b30904252e72b4f245fde

Btw, clawback does not reset epoch. Epoch is only updated in the newly added transactions.

XLS-draft-confidential-mpt-key-rotation/README.md
yinyiqian1: https://github.com/XRPLF/XRPL-Standards/pull/598/changes/c1734cbcac1ea305793b69bc6756acd5a0bcc3a5

XLS-draft-confidential-mpt-key-rotation/README.md
yinyiqian1: moved

XLS-draft-confidential-mpt-key-rotation/README.md
yinyiqian1: https://github.com/XRPLF/XRPL-Standards/pull/598/changes/c27e09f559d388d162942fb131d43d73b65a94fb

XLS-draft-confidential-mpt-key-rotation/README.md
yinyiqian1: https://github.com/XRPLF/XRPL-Standards/pull/598/changes/6b2ebd54b1fbd079e275642946ed6ee25c9b3642

XLS-draft-confidential-mpt-key-rotation/README.md
yinyiqian1: https://github.com/XRPLF/XRPL-Standards/pull/598/changes/698d1e5280b0f6db968791a45d620c4919236daf

XLS-draft-confidential-mpt-key-rotation/README.md
yinyiqian1: https://github.com/XRPLF/XRPL-Standards/pull/598/changes/1c876009cfff616fb15f3262a94cdc1272ab7bfa

XLS-draft-confidential-mpt-key-rotation/README.md
yinyiqian1: https://github.com/XRPLF/XRPL-Standards/pull/598/changes/fab24532322aa0eea896002b42c531febb186878

XLS-0099-confidential-mpt-key-rotation/README.md
yinyiqian1: https://github.com/XRPLF/XRPL-Standards/pull/598/changes/13071275fac3ed552f0b08715a2a899764ee2fad

XLS-draft-confidential-mpt-key-rotation/README.md
yinyiqian1: https://github.com/XRPLF/XRPL-Standards/pull/598/changes/ee5364108fc0c52527fee3c073a80f835f995fd7

XLS-draft-confidential-mpt-key-rotation/README.md
yinyiqian1: https://github.com/XRPLF/XRPL-Standards/pull/598/changes/6ae3c6192853308da0b5c490b1c6b05ab084c35d

XLS-draft-confidential-mpt-key-rotation/README.md
```diff
@@ -0,0 +1,1301 @@
...
+
+#### 4.6.1. Determining Mirror Status
+
+For an `MPToken` with initialized confidential state:
+
+- The issuer mirror is current when `IssuerEncryptedBalance` is present and `IssuerKeyMirrorEpoch` equals `IssuerKeyEpoch`.
+- When an auditor key is configured, the auditor mirror is current when `AuditorEncryptedBalance` is present and `AuditorKeyMirrorEpoch` equals `AuditorKeyEpoch`. When no auditor key is configured, no auditor mirror is required.
+- Migration is required whenever a required mirror is not current.
+
+An absent epoch field is treated as epoch 0. Validators determine whether a mirror is current by requiring equality between its mirror epoch and the corresponding key epoch. Ledger invariants separately ensure that a mirror epoch cannot exceed its corresponding key epoch.
```
mDuo13: In general, don't use "validators" in cases like these where you actually mean "everyone in the network"; we don't want to layer excess responsibilities on validators.

```suggestion
An absent epoch field is treated as epoch 0. A mirror field is current if its mirror epoch and the corresponding key epoch are equal. Ledger invariants separately ensure that a mirror epoch cannot exceed its corresponding key epoch.
```

XLS-draft-confidential-mpt-key-rotation/README.md
```diff
@@ -0,0 +1,1301 @@
...
+
+### 4.5. Re-encryption Strategy
+
+Active re-encryption is the recommended strategy. The issuer submits `ConfidentialMPTMirrorUpdate` for all holders after rotating the key. The issuer is recommended to lock each holder's `MPToken` before submitting `ConfidentialMPTMirrorUpdate` and unlock after success. Note: this is a recommendation, not a requirement. The epoch staleness check already prevents the holder from successfully transacting with a stale mirror - the lock only prevents the holder from wasting fees on transactions that will be rejected. However, lock + `ConfidentialMPTMirrorUpdate` + unlock = 3 transactions per holder, tripling migration transaction volume at large holder counts. Issuers should weigh this cost against the UX benefit.
```
mDuo13: I don't understand how the three-step lock/update/unlock process prevents holders from wasting fees.

If a holder's transaction is rejected for being stale, it fails with a `tec` code, which burns the XRP transaction cost but pays no other fees. If that transaction is instead rejected because the MPT is locked, it fails with a different `tec` code which similarly burns the transaction cost only. Am I missing something here?

XLS-draft-confidential-mpt-key-rotation/README.md
```diff
@@ -0,0 +1,1301 @@
...
+- Issuer is passive for issuer mirrors - only the holder can act
+- Holders know b from decrypting `ConfidentialBalanceSpending` via sk_H and re-encrypt it under the new pk_I' using the cross-key equality proof (Section 5.4.6)
+
+This is a one-step process per holder - unlike holder key loss recovery which requires two steps (holder authorizes, issuer completes). Here the holder acts alone with no issuer involvement needed for their mirror.
+
+**How it works**:
+
+1. Issuer registers new pk_I' via `MPTokenIssuanceSet` (they still have their XRPL signing key).
+2. `IssuerKeyEpoch` increments.
+3. Validators reject holder confidential transactions that require an issuer mirror while `IssuerKeyMirrorEpoch` does not equal `IssuerKeyEpoch`.
```
mDuo13: As above, avoid language like this that puts the responsibility of rejecting transactions solely on validators. Every node in the network processes transactions the same way, regardless of validator or not. Also, the word "rejects" is often used specifically for not allowing transactions into the consensus set, but to my understanding, these are accepted but fail with a `tec` code, so the word "fail" is more appropriate here. (You could optionally say "fail with an error code" or maybe even add the specific error code.)

```suggestion
3. Holder confidential transactions that require an issuer mirror fail if `IssuerKeyMirrorEpoch` does not equal `IssuerKeyEpoch`.
```

XLS-draft-confidential-mpt-key-rotation/README.md
```diff
@@ -0,0 +1,1301 @@
...
+
+Consistent with XLS-0096's existing behavior for initial key setting. A rogue issuer key only harms the issuer's own mirror and clawback capability.
+
+### 12.5. `ConfidentialBalanceVersion` Increment on Holder Key Rotation
+
+`ConfidentialMPTHolderKeyUpdate` in rotation mode increments `ConfidentialBalanceVersion`, invalidating in-flight `ConfidentialMPTSend` proofs. See Section 10.2.
+
+### 12.6. Two-Step Recovery Authorization
+
+The holder's authorization (`tfHolderKeyRecovery`) is signed by the holder's XRPL signing key. Validators enforce that `RecoveryKey` is present before accepting `ConfidentialMPTRecoverBalance`. The issuer cannot act unilaterally.
```
mDuo13: As above, avoid language that suggests validators are the only ones enforcing the rules.

```suggestion
The holder's authorization (`tfHolderKeyRecovery`) is signed by the holder's XRPL signing key. The `ConfidentialMPTRecoverBalance` transaction must include a `RecoveryKey` field. The issuer cannot act unilaterally.
```

XLS-draft-confidential-mpt-key-rotation/README.md
```diff
@@ -0,0 +1,1301 @@
...
+
+Historical issuer secret-key retention is optional and depends on the issuer's migration strategy. If the issuer intends to actively migrate any holder whose `IssuerEncryptedBalance` remains encrypted under an older issuer key, the issuer MUST retain the corresponding historical secret key until those holder mirrors have been migrated.
+
+Once an issuance-wide holder traversal confirms that no issuer mirror remains at the corresponding epoch, the historical secret key is no longer required and may be destroyed. An issuer may instead destroy the old secret key earlier and rely on the remaining holders to self-migrate using `sk_H`. Doing so does not affect ledger correctness, but permanently removes the issuer-driven migration path for those holders.
+
+Historical auditor secret keys are not required for mirror migration because auditor mirrors are reconstructed from a current issuer mirror or through holder self-migration.
+
+### 12.10. Holder Self-Migration Security
+
+For holder self-migration (Section 5.4.6), validators must enforce:
```
mDuo13:
```suggestion
Holder self-migration (Section 5.4.6) requires all of the following:
```

XLS-draft-confidential-mpt-key-rotation/README.md
```diff
@@ -0,0 +1,1301 @@
...
+
+Until a holder's mirror is migrated, confidential transactions for that holder are rejected. The underlying reason is cryptographic: in `ConfidentialMPTSend` and other transactions, the holder correctly reads the current `IssuerEncryptionKey` (`pk_I'`) from the issuance and constructs `IssuerEncryptedAmount` under it. However, the on-ledger accumulated `IssuerEncryptedBalance` is still under the old key (`pk_I`). Adding a delta under `pk_I'` to an accumulated balance under `pk_I` is cryptographically invalid - ciphertexts under different keys cannot be combined homomorphically.
+
+Comparing the holder's mirror epoch with the corresponding issuance key epoch makes this detectable cleanly:
+
+- Validators use it to reject the transaction explicitly and early, before attempting an invalid homomorphic addition
```
mDuo13:
```suggestion
- Servers in the network use it to reject the transaction explicitly and early, before attempting an invalid homomorphic addition
```

mDuo13 (commented):
Noted a couple language issues and an open question re: the suggested migration process.

I also have some concerns that the issuer key migration process is cumbersome at scale. It seems like a problem that it stops users from spending using their confidential balances and I worry issuers may neglect to migrate individual holders either by accident or intention. As I understand it, holders can proactively self-migrate, so this is more of a nuisance than an outright freeze on the funds, but I'd be supportive of a more gradual key migration path if possible.

XLS-draft-confidential-mpt-key-rotation/README.md
yinyiqian1: The transaction fee for `ConfidentialMPTMirrorUpdate` is 10*baseFee because it requires the expensive zero knowledge proof verification. So failing `ConfidentialMPTMirrorUpdate` is more expensive than lock.
I'll update the wording here to clarify this, as the current phrasing might be confusing.

XLS-draft-confidential-mpt-key-rotation/README.md
yinyiqian1: Modified the `Validator` wording in https://github.com/XRPLF/XRPL-Standards/pull/598/changes/cc5c08e90222e8e1d0a63dd8b7c26b864ac475f9

XLS-draft-confidential-mpt-key-rotation/README.md
yinyiqian1: Modified the `Validator` wording in https://github.com/XRPLF/XRPL-Standards/pull/598/changes/cc5c08e90222e8e1d0a63dd8b7c26b864ac475f9

XLS-draft-confidential-mpt-key-rotation/README.md
yinyiqian1: https://github.com/XRPLF/XRPL-Standards/pull/598/changes/cc5c08e90222e8e1d0a63dd8b7c26b864ac475f9

XLS-draft-confidential-mpt-key-rotation/README.md
yinyiqian1: Modified the `Validator` wording in https://github.com/XRPLF/XRPL-Standards/pull/598/changes/cc5c08e90222e8e1d0a63dd8b7c26b864ac475f9

XLS-draft-confidential-mpt-key-rotation/README.md
yinyiqian1: Good catch. Lock is not needed regarding the fee. The cost is the same. I removed this paragraph https://github.com/XRPLF/XRPL-Standards/pull/598/changes/4077d6e5d90e9eb7caefe8404b206e47dc5d70be

mDuo13 (approved):
:+1: Changes look good.

Tapanito (approved):

Tapanito (approved):

PR #603, opened 2026-08-17
Design transaction_proposal RPC around per-account signing status

ckeshava:
## Summary

Redesigns the `transaction_proposal` RPC (§8.1) of XLS-0103 so clients get a computed, per-account view of how far a proposal is from a submittable transaction, instead of a single flat `signed_weight`/`quorum`/`status` triple:

- Renames the request field `ticket` to `ticket_seq`, aligning it with the addressing `ledger_entry` accepts for a `transaction_proposal` object (identical field types and error codes for malformed addressing).
- Replaces the flat response with `proposal_status` (`pending`/`complete`/`expired`, with `expired_reason`) plus a `signing_status` array: one row per required authorization, keyed by (`account`, `role`), with `signed`, `reason`, and per-signer `SignerList` detail so a wallet knows exactly whom to chase next. Adds `tx_blob` on `complete`.
- Adds §8.1.3 (Completeness Evaluation): how required authorizations are derived (`account`, `counterparty`, `sponsor`, `batch_participant`), authorization-vs-cryptography semantics against the live ledger, expiry bounds, and what `complete` does and does not guarantee (point-in-time verdict, ledger-freshness caveats).
- Expands the failure conditions and adds a proposed-`Batch` example response — the motivating case for the per-account design.
- Resolves the "Completion signalling" open question in §12 (now answered by this design).
- Corrects §13.5's stale-signer semantics: removing a collected signer does not merely subtract weight — submission rejects the entire collected set (`tefBAD_SIGNATURE`), making the proposal permanently unsatisfiable, so it must be cancelled and re-proposed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/XLS-0103-onchain-cosigner/README.md
+++ b/XLS-0103-onchain-cosigner/README.md
@@ -869,8 +869,7 @@ Rows are keyed by the pair (`account`, `role`), not by `account` alone. The same
 
 | Value                            | Meaning                                                                                                                          |
 | -------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------- |
-| `no_signature`                   | No signature material has been collected for this row yet.                                                                          |
-| `below_quorum`                   | A `Signers` array is collected but its weight against the live `SignerList` is below the live `SignerQuorum`.                       |
+| `inadequate_signatures` | `bool` | In the case of Single-Sign configuration, the account has not received any signatures. In the case of Multi-Sign configuration, the received signatures do not yet satisfy the requisite quorum |
 | `invalid_signer_set`             | The collected `Signers` array contains an entry the live `SignerList` does not authorize; submission rejects the set wholesale. This occurs when the Account has removed a Signer from its erstwhile SignerList configuration    |
 | `no_signer_list`                 | A `Signers` array is collected but the account has no `SignerList` on the queried ledger. This error occurs when the Account had a SignerList but has since deleted it and switched to a SingleSign configuration                                          |
 | `master_disabled`                | The collected signature is by the master key, and the master key has since been disabled. This is caused by the updated ledger-state since the collection of this signature.                                           |
--- a/XLS-0103-onchain-cosigner/README.md
+++ b/XLS-0103-onchain-cosigner/README.md
@@ -911,7 +911,7 @@ An inner transaction initiated by the outer account adds no row: the outer accou
 Each signature was cryptographically verified when `TransactionProposalSign` appended it (§6.3.2), and ledger data cannot change after that. What *can* change is whether the signature still **authorizes** the account, so `signed` is computed by re-applying the standard authorization rules against the queried ledger — the same rules submission applies:
 
 - A single signature must be by the account's current regular key, or by its master key while the master key is enabled.
-- A collected `Signers` array is scored against the account's **live** `SignerList`: entries no longer on the list contribute nothing, an entry the list does not authorize invalidates the whole set (as it would at submission), and the surviving weight must meet the live `SignerQuorum`.
+- A collected `Signers` array is validated against the account's **live** `SignerList`, and it fails **as a whole** if *any* collected entry is not currently authorized: an entry absent from the list, one whose master key has since been disabled, one whose regular key has since rotated, or a non-phantom entry whose account no longer exists each void the entire set at submission (`tefBAD_SIGNATURE` / `tefMASTER_DISABLED`). Otherwise the set's weight must meet the live `SignerQuorum`. Because signatures only ever accumulate on a proposal, a voided set cannot be repaired by further signing — the practical remedy is `TransactionProposalCancel` and a fresh proposal.
 - A `BatchSigners` entry for an account that does not exist yet is authorized only by that account's own master key (an earlier inner transaction may create the account).
 - A sponsor row is satisfied by a valid `SponsorSignature`, or — only while no `SponsorSignature` field has been collected at all — by an on-ledger `Sponsorship` entry between the sponsor and the initiator whose flags do not require a co-signature for what this transaction sponsors. A collected-but-no-longer-authorizing `SponsorSignature` is **not** rescued by the exemption, because submission validates a present `SponsorSignature` unconditionally.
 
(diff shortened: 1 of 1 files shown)


XLS-0103-onchain-cosigner/README.md
```diff
@@ -824,57 +824,141 @@ To use a proposal, a signer or wallet has to fetch it and see how far along it i
...
+The response returns the raw ledger object plus **computed convenience fields** so a client does not have to join the collected signatures against live `SignerList`s itself:
+
+| Field             | Type   | Description                                                                                                       |
+| ----------------- | ------ | ------------------------------------------------------------------------------------------------------------------ |
+| `proposal_id`     | string | The ID of the `TransactionProposal`.                                                                                |
+| `proposal`        | object | The raw `TransactionProposal` ledger object.                                                                        |
+| `proposal_status` | string | Where the proposal is in its lifecycle: `"pending"`, `"complete"`, or `"expired"` (see below).                      |
+| `expired_reason`  | string | Present only when `proposal_status` is `"expired"`: `"expiration"` or `"last_ledger_sequence"` (§8.1.3.3).          |
```
shawnxie999: do we need this field? i would assume normal users would only care whether it has expired, not how it expired

XLS-0103-onchain-cosigner/README.md
```diff
@@ -824,57 +824,141 @@ To use a proposal, a signer or wallet has to fetch it and see how far along it i
...
+| `tx_blob`         | string | Present only when `proposal_status` is `"complete"`: the stored `ProposedTransaction` serialized in submit-ready binary form, so a client does not have to reassemble and re-serialize it. Providing it implies nothing beyond §8.1.3.4. |
+
+> The field is named `proposal_status`, not `status`, because `status` is already the RPC envelope's success/error indicator and the two would collide in the same JSON object.
+
+Each `signing_status` entry:
+
+| Field           | Type    | Description                                                                                                          |
+| --------------- | ------- | ---------------------------------------------------------------------------------------------------------------------- |
+| `account`       | string  | The account whose authorization is required.                                                                            |
+| `role`          | string  | Why it is required: `"account"`, `"batch_participant"`, `"counterparty"`, or `"sponsor"` (§8.1.3.1).                    |
```
shawnxie999: The  `role`  field feels ambiguous and may not be necessary. In a batch, the counterparty and sponsor are also identifiable as batch participants. So we might want to remove it.

XLS-0103-onchain-cosigner/README.md
```diff
@@ -824,57 +824,141 @@ To use a proposal, a signer or wallet has to fetch it and see how far along it i
...
+
+Rows are keyed by the pair (`account`, `role`), not by `account` alone. The same account can owe two independent authorizations through different signature slots — for example, the fee sponsor of a proposed `Batch` who is also an inner participant signs once through `SponsorSignature` and once through `BatchSigners` — and each slot succeeds or fails on its own.
+
+`reason` values:
+
+| Value                            | Meaning                                                                                                                          |
+| -------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------- |
+| `no_signature`                   | No signature material has been collected for this row yet.                                                                          |
```
shawnxie999: this seems redundant, `below_quorum` can already cover the case of no signature.

XLS-0103-onchain-cosigner/README.md
```diff
@@ -824,57 +824,141 @@ To use a proposal, a signer or wallet has to fetch it and see how far along it i
...
+| `below_quorum`                   | A `Signers` array is collected but its weight against the live `SignerList` is below the live `SignerQuorum`.                       |
+| `invalid_signer_set`             | The collected `Signers` array contains an entry the live `SignerList` does not authorize; submission rejects the set wholesale.     |
```
shawnxie999: Can add a bit in the description that it's caused by having the target account removing signers from their SignersList

XLS-0103-onchain-cosigner/README.md
```diff
@@ -824,57 +824,141 @@ To use a proposal, a signer or wallet has to fetch it and see how far along it i
...
+| `signed`        | boolean | Whether the signature material collected so far currently authorizes this account on the queried ledger (§8.1.3.2).     |
+| `reason`        | string  | Present only when `signed` is `false`: why (see below).                                                                 |
+| `signed_weight` | number  | Present only when a `Signers` array has been collected for this row: its weight against the account's live `SignerList`. |
+| `quorum`        | number  | Present only when the account has a live `SignerList`: its `SignerQuorum`.                                              |
+| `signers`       | array   | Present only when the account has a live `SignerList`: one entry per list member — `{account, weight, signed}` — where `signed` is whether a currently-valid signature from that member has been collected. This is the list a wallet chases: every member with `signed: false` is a candidate next signer. Collected signatures from accounts *not* on the live list do not appear here; they surface as `reason: "invalid_signer_set"`. |
```
shawnxie999: what happens if it is a single signed transaction? then in that case, quorum won't apply. We may probably want to make these fields optional for the exception when the transaction has been singled signed.

mvadari:
@ckeshava please don't use "XLS-103" in your PR title, that hasn't been assigned yet.

XLS-0103-onchain-cosigner/README.md
ckeshava: Yes, in the case of a single-sign account (RegularKey / Masterkey), the fields `quorum`, `signer_weight` and `signers` does not apply. These are `optional` fields.

XLS-0103-onchain-cosigner/README.md
ckeshava: Yes, this is true. Futhermore, one account could fill in multiple roles (Counterparty and a batch-inner-transaction), which makes the role field multi-value field. I intended this as a information field only.

I will remove it 👍

XLS-0103-onchain-cosigner/README.md
ckeshava: I don't have strong opinions on this, this is a good-to-have convinience field. We can add it later if required 👍 

I will remove it from the spec.

XLS-0103-onchain-cosigner/README.md
ckeshava: It is a tad confusing for users if we include `below_quorum` field for Accounts with a Single-Signature scheme (Regularkey / Masterkey), is it not?

If that is acceptable, then I will remove this field.

The justification for having both the fields is to differentiate the "pending" transaction-proposal when the account-in-question has a SignerList versus Single-Signing configuration.

Alternatively, we can have one single-field and rename it to:
```
| `inadequate_signatures` | `bool` | In the case of Single-Sign configuration, the account has not received any signatures. In the case of Multi-Sign configuration, the recieved signatures do not yet satisfy the requisite quorum |
```

What would you suggest?

XLS-0103-onchain-cosigner/README.md
ckeshava: updated this in 06d1645

XLS-0103-onchain-cosigner/README.md
ckeshava: updated this in 06d1645

XLS-0103-onchain-cosigner/README.md
ckeshava: updated this in 06d1645

XLS-0103-onchain-cosigner/README.md
ckeshava: added more description in 06d1645

ckeshava:
Note to reviewers: Prettier and the Github-CI checks necessitate ~70 lines of non-relevant changes to this PR. I can do this after the reviewers are satisfied with the semantic and textual diff in this PR. This ensures a minimal diff.

XLS-0103-onchain-cosigner/README.md
shawnxie999: Makes sense, I agree `inadequate_signatures` is a good name

XLS-0103-onchain-cosigner/README.md
shawnxie999: Sounds good, these fields are easy add over time, so we can always add it if there's the need

XLS-0103-onchain-cosigner/README.md
ckeshava: ok. Specified the new field in this diff: 06d1645..6157ea9

shawnxie999 (approved):

ckeshava:
@shawnxie999  I do not see a "Merge" or a "Squash and Merge" button in my Github UI. Let me know if there is anything that needs to be done from my end.

shawnxie999:
@ckeshava just merged

Issue #10, opened 2020-03-31
XLS-9d: Blinded Tags


nbougalis:
<pre>
  XRP Ledger Proposed Standard #9

  Title:        Blinded Tags
  Author:       Nikolaos D. Bougalis <nikb@bougalis.net>
  Affiliation:  Ripple
  Created:      2020-03-30
</pre>

# Abstract

The XRP Ledger supports source and destination tags which can allow a single account to hold funds for multiple users. These arrangements are commonly called _hosted wallets_. When sending _from_ a hosted wallet, the account uses a **source tag** to tell which user prompted the transaction. When sending _to_ a hosted wallet, the **destination tag** tells the account how to further distribute those funds.

A typical example is an exchange like Bitstamp. When you want to deposit funds, Bitstamp provides two pieces of information: an address identifying a Bitstamp wallet, and a destination tag. Several (if not all) Bitstamp customers use the same Bitstamp address, but each customer gets a separate, unique destination tag.

A destination tag is presently an unsigned 32-bit integer. This means that there are 4,294,967,296 possible destination tags. While that's certainly enough tags to allow an exchange to use a unique tag per customer today and perhaps even for the foreseeable future, using a fixed tag presents privacy challenges: it allows transactions to be correlated by treating the `{ address, tag }` pair as a unique address corresponding to a single customer.

As a possible workaround, exchanges could allow users to generate new tags as necessary. In fact, several existing exchanges do just that. The problem with this scheme is that it complicates things for the exchange and for the customer. This also uses more destination tags, which are still finite.

This proposal attempts to alleviate this concern by specifying a standard that allows for _tag blinding_. A blinded tag is mutated in such a way that it is meaningful only to the sender and the recipient of a transaction, but appears random to everyone else.

# The Proposal

The goal of this proposal is to ensure that if blinded tags are in use, an attacker capable of observing every payment transaction will be unable to isolate a pair of transactions that refer to the same unblinded tag. This proposal aims to be secure, minimal, and performant; ideally, it should be possible to implement tag blinding as a single function call that does not noticeably increase the time necessary to assemble a transaction. Similarly, using a blinded tag should not make it significantly harder for the intended recipient to process a transaction.

## Protocol Changes

This proposal assumes several new flags and fields for supporting blinded destination tags. One amendment to the XRP Ledger would introduce all of the following:

Two new [AccountRoot ledger-state flags](https://xrpl.org/accountroot.html#accountroot-flags):

[...]

PR #613, opened 2026-08-26
Onchain Cosigner

shawnxie999:
### Type of Change

- [ ] New XLS Draft


XLS-onchain-cosigner/README.md
```diff
@@ -0,0 +1,1177 @@
...
+
+## 3. Overview
+
+### 3.1. Terminology
+
+- **Proposal**: A `TransactionProposal` ledger object. It holds a single unsigned **proposed transaction** (the payload) and the set of signatures collected for it so far.
+- **Proposed transaction**: The transaction that will be executed on behalf of the **target account** once enough signatures are collected. It is stored, immutable, inside the proposal. (This is a distinct concept from a Batch "inner transaction".)
+- **Target account**: The account on whose behalf the proposed transaction executes — i.e. the `Account` of the proposed transaction. Its `SignerList` configuration governs the quorum.
+- **Proposer**: The account that submits `TransactionProposalCreate`. It owns the proposal object and pays its reserve. The proposer need not be a signer or the target account.
+- **Signer**: An account ID on the target account's applicable `SignerList` that can append its signature to the proposal, contributing its weight toward quorum. For multi-signing, this may be an unfunded AccountID derived from a public key, matching existing XRPL multi-sign behavior.
```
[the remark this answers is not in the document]
shawnxie999: https://github.com/XRPLF/XRPL-Standards/pull/613/commits/7f67520609ede2bbae578c0fce2d6f0620800429

XLS-draft-onchain-cosigner/README.md
```diff
@@ -0,0 +1,1177 @@
+<pre>
+  title: On-Chain Cosigner
+  description: Native on-ledger proposal and multi-signature collection for XRPL transactions.
+  author: Shawn Xie, Zhiyuan Wang, Chenna Keshava B S, Mayukha Vadari
+  category: Amendment
+  status: Draft
+  created: 2026-07-14
```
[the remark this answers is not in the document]
shawnxie999: https://github.com/XRPLF/XRPL-Standards/pull/613/commits/7f67520609ede2bbae578c0fce2d6f0620800429

XLS-onchain-cosigner/README.md
```diff
@@ -0,0 +1,1177 @@
...
+      "...": "..."
+    }
+  },
+  "proposal_status": "pending",
+  "signing_status": [
+    {
+      "account": "rTARGET..........................",
+      "role": "account",
+      "signed": false,
+      "reason": "below_quorum",
```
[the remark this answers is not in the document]
shawnxie999: https://github.com/XRPLF/XRPL-Standards/pull/613/commits/7f67520609ede2bbae578c0fce2d6f0620800429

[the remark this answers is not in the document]
shawnxie999: https://github.com/XRPLF/XRPL-Standards/pull/613/commits/7f67520609ede2bbae578c0fce2d6f0620800429

XLS-draft-onchain-cosigner/README.md
```diff
@@ -0,0 +1,1186 @@
...
+
+#### 6.3.2. Protocol-Level Failures
+
+1. No `TransactionProposal` object exists with the given `ProposalID` (`tecNO_ENTRY`).
+2. The proposal is terminal — its `Expiration` has passed, or the proposed transaction's `LastLedgerSequence` has passed (`tecEXPIRED`). This is a claimed-fee failure: no signature is recorded, but the terminal proposal is deleted as a side effect (see §6.4). This condition is checked before the authorization conditions below.
+3. `SigningFor` is not an account the proposed transaction requires a signature from — it is not the transaction's `Account`/`Delegate`, its `Counterparty`, its `Sponsor`, or (for a `Batch`) an account owning an inner transaction in `RawTransactions` (`tecNO_PERMISSION`).
+4. The signer is not authorized: for single-signing, `ProposalSignature.SigningPubKey` is not `SigningFor`'s master or regular key; for multi-signing, `ProposalSignature.Account` is not on `SigningFor`'s applicable `SignerList`, or `ProposalSignature.SigningPubKey` is not valid for `ProposalSignature.Account` under standard multi-sign rules (`tecNO_PERMISSION`).
```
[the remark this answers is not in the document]
shawnxie999: XLS49 is not a feature yet.

XLS-draft-onchain-cosigner/README.md
```diff
@@ -0,0 +1,1186 @@
...
+- **Must** specify a `TicketSequence` for its target account and **must not** specify `Sequence`. Requiring a ticket decouples the proposed transaction from the target account's live sequence, so unrelated target-account activity cannot invalidate the proposal while signatures are being collected (see §9.2). For as long as the proposal exists, that ticket is reserved: a transaction may consume it only if its payload matches `ProposedTransaction`; otherwise, it is rejected.
+- **Must** carry a `Fee`. The proposed transaction's fee is paid by the **target account** (the proposed transaction's `Account`) when the completed transaction is submitted.
```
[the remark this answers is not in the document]
shawnxie999: Should add that in TransactionProposalSign, for batch, it checks that with each additional signer, the fee is still sufficient.

XLS-draft-onchain-cosigner/README.md
```diff
@@ -0,0 +1,1186 @@
...
+- **`Batch` proposed transaction:** 10 owner-reserve increments (currently **2 XRP**), reflecting its larger footprint (up to 8 inner transactions and signatures for multiple participant accounts).
+
+Each increment is the standard owner-reserve amount (currently 0.2 XRP, subject to Fee Voting).
+
+### 4.5. Deletion
+
+**Terminal proposal:** A proposal is **terminal** when it can no longer be completed and submitted, i.e. when either of the following is true relative to the parent ledger:
+
+- The parent ledger's close time is at or after `Expiration`; or
+- The proposed transaction includes a `LastLedgerSequence` and the current ledger sequence is greater than it.
```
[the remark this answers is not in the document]
shawnxie999: valid suggestion, a terminal proposal can still be submitted, even if the proposal past expiration - it only means it will no longer accept sigs

mvadari:
@ai-review --aggressive

XLS-draft-onchain-cosigner/README.md
```diff
@@ -0,0 +1,1186 @@
...
+6. A `TransactionProposal` with the same `ProposalID` already exists — i.e. a live proposal (owned by anyone) already targets the same account with the same `TicketSequence` (`tecDUPLICATE`, §4.1).
+7. The proposed transaction's `TicketSequence` is not a valid `Ticket` of the target account (`tefNO_TICKET`).
+8. The proposer is neither the target account — or the `Delegate`, when one is present — nor a member of that account's applicable `SignerList` (`tecNO_PERMISSION`).
+
+### 5.4. State Changes
+
+**On Success (`tesSUCCESS`):**
+
+- Creates a new `TransactionProposal` ledger object whose `Owner` is the sending `Account`. The proposed transaction is stored with no signatures yet.
+- Increments the `Owner`'s `OwnerCount`.
```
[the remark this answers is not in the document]
shawnxie999: Yes, should be incremented with 5/10

XLS-draft-onchain-cosigner/README.md
```diff
@@ -0,0 +1,1186 @@
...
+If `SigningFor` does not match one of these required accounts, the transaction fails with `tecNO_PERMISSION`. When it does match, the ledger records the signature in the location that corresponds to that account's role:
+
+| If `SigningFor` is the transaction's… | The signature is recorded in…                            |
+| ------------------------------------- | -------------------------------------------------------- |
+| `Account` / `Delegate`                | `ProposedTransaction.Signers` (or top-level, see §6.1.2) |
+| `Counterparty`                        | `ProposedTransaction.CounterpartySignature`              |
+| `Sponsor`                             | `ProposedTransaction.SponsorSignature`                   |
+| `Batch` participant                   | `ProposedTransaction.BatchSigners[SigningFor]`           |
+
+If the same account fills **more than one** role, such as being both the `Counterparty` and the `Sponsor`, the same contribution is recorded in **every** matching slot. Each slot is still validated independently. In most cases these roles are different accounts, so one contribution fills one slot.
```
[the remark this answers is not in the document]
shawnxie999: It's fine, in that case, an inner tx with SponsorSignature is just treated as a BatchSigner

XLS-draft-onchain-cosigner/README.md
```diff
@@ -0,0 +1,1186 @@
...
+| --------------- | ------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `account`       | string  | The account whose authorization is required.                                                                                                                                                                                                                                                                                                                                                                                                                                                   |
```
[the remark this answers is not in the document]
shawnxie999: Role is actually removed

XLS-draft-onchain-cosigner/README.md
```diff
@@ -0,0 +1,1186 @@
...
+
+### 13.4. Cancellation does not revoke already-collected signatures
+
+This is the central security consideration of the copy-and-submit model. The proposal object is a bulletin board, not an execution gate: once a quorum-weight of valid signatures has been collected, any observer may have copied them, and those signatures remain valid regardless of whether the proposal object still exists. Cancelling or expiring the proposal frees the reserve but does **not** guarantee the transaction will not execute. To positively prevent execution of a completed (or nearly-completed) proposal, the target account must consume the proposed transaction's `TicketSequence` with another transaction and/or rely on its `LastLedgerSequence` window elapsing. Architects and wallets should surface this clearly.
```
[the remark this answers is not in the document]
shawnxie999: once the proposal object is deleted, the ticket is no longer reserved. so yes, there might be a race, but that's fine.

XLS-draft-onchain-cosigner/README.md
```diff
@@ -0,0 +1,1186 @@
...
+
+### 13.6. Denial-of-service and reserve pressure
+
+Each proposal consumes an elevated flat owner reserve (§4.4) held against the `Owner` — higher than a typical ledger entry, and higher still for a `Batch` — pricing the larger state burden and disincentivizing spam. Because every appended signature must be valid, an attacker cannot inflate a proposal with junk. Built-in expiry ensures abandoned proposals can always be cleaned up (by anyone, once terminal) so they do not accumulate indefinitely in ledger state.
+
+Because anyone may propose against any account and there is one slot per `(target account, ticket)` (§4.1), an attacker could **squat** a slot the real proposer wanted, blocking it with `tecDUPLICATE`. Each attempt costs a full reserve, and tickets give the honest proposer far more slots than an attacker could block. The target account is also never stuck with an unwanted proposal: it may delete any proposal made for it via `TransactionProposalCancel` at any time (§7.2), clearing the slot and the reserved ticket regardless of who created the proposal or how far along it is.
```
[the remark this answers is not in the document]
shawnxie999: yes you are right

XLS-draft-onchain-cosigner/README.md
```diff
@@ -0,0 +1,1186 @@
...
+3. The proposed transaction is not unsigned. The result depends on the signature field and its contents:
+   - A non-empty `TxnSignature`; a non-empty `SigningPubKey` combined with a `Signers` array; or `Signers` entries carrying real signatures returns `temINVALID`.
+   - A non-empty `SigningPubKey` that is not a valid key type returns `temBAD_SIGNATURE`.
+   - A proposed `LoanSet` whose `CounterpartySignature` holds a real signature returns `temINVALID`.
+   - A `SponsorSignature` without the corresponding `Sponsor` and `SponsorFlags` fields returns `temMALFORMED`.
```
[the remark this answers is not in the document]
shawnxie999: that should be checked in preflight already.

XLS-draft-onchain-cosigner/README.md
```diff
@@ -0,0 +1,1186 @@
...
+- **Must** specify a `TicketSequence` for its target account and **must not** specify `Sequence`. Requiring a ticket decouples the proposed transaction from the target account's live sequence, so unrelated target-account activity cannot invalidate the proposal while signatures are being collected (see §9.2). For as long as the proposal exists, that ticket is reserved: a transaction may consume it only if its payload matches `ProposedTransaction`; otherwise, it is rejected.
```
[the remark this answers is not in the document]
shawnxie999: yes the code will be terTICKET_RESERVED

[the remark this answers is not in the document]
ckeshava: This observation is correct. We could improve the clarity of Section-11, by specifying the error code.

XLS-draft-onchain-cosigner/README.md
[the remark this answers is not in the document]
ckeshava: - The specified fees could be inadequate due to high network traffic. It is not possible to determine this at the time of proposal creation.
- The case of insufficient fees is also possible in transactions like LoanSet (which could have up to two signatures) and transactions with Sponsor-signature.

XLS-draft-onchain-cosigner/README.md
```diff
@@ -0,0 +1,1186 @@
...
+
+#### 6.1.1. `SigningFor` — which account is being authorized, and where the signature lands
+
+`SigningFor` answers the question: **whose approval is this signature providing?** It must name an account that the **proposed transaction** needs a signature from. In other words, `SigningFor` must point to one of the proposed transaction's signature slots:
+
+- the proposed transaction's own `Account`, or its `Delegate` if permission delegation is used;
+- the `Counterparty`, if that transaction type has one (for example, a [`LoanSet` (XLS-66)](../XLS-0066-lending-protocol/README.md) lender; if omitted, this defaults to the `LoanBroker.Owner`, XLS-66 §3.8);
+- the `Sponsor`, if the transaction is sponsored ([XLS-68](../XLS-0068-sponsored-fees-and-reserves/README.md));
+- for a `Batch`, any account whose signature is needed for the batch, including each inner transaction's account and any additional account required by an inner transaction, such as its `Delegate`, `Counterparty`, or `Sponsor`. Each of these accounts is a batch **participant**.
```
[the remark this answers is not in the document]
ckeshava: Reminder: Did you mean to accept this suggestion? @shawnxie999

XLS-draft-onchain-cosigner/README.md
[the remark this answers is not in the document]
ckeshava: Yes, I agree that the specification of role is not necessary. An account could fulfill multiple roles (Batch-Inner-Signer and SponsorSigner). Both of these roles could be satisfied by a collection of its signatures.

XLS-draft-onchain-cosigner/README.md
```diff
@@ -0,0 +1,1186 @@
...
+
+### 6.3. Failure Conditions
+
+#### 6.3.1. Data Verification
+
+All Data Verification failures return a `tem`-level error.
+
+1. `ProposalID` is missing or malformed (`temMALFORMED`).
+2. `SigningFor`, `ProposalSignature`, `ProposalSignature.Account`, `ProposalSignature.SigningPubKey`, or `ProposalSignature.TxnSignature` is missing (`temMALFORMED`).
+3. `ProposalSignature.TxnSignature` is not valid over `SigningFor`'s signing data for the proposed transaction (§6.1.2) (`temBAD_SIGNATURE`).
```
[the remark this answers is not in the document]
ckeshava: This is a valid suggestion ^

XLS-draft-onchain-cosigner/README.md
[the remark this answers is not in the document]
ckeshava: My bad, I forgot to remove all the occurences of Role field. Do you want me to create a new PR with the removed-Role instances?

XLS-draft-onchain-cosigner/README.md
```diff
@@ -0,0 +1,1186 @@
...
+- **Multiple Signer Lists (XLS-49):** Per-transaction-type signer lists are honored automatically, since both the per-signature check and the final submission use standard multi-sign resolution for the proposed transaction's type.
+
+## 11. Backwards Compatibility
+
+This proposal is purely additive: it introduces one new ledger entry type and three new transaction types, all gated behind the `Cosigner` amendment. Existing multi-sign, `SignerListSet`, and off-chain signing workflows are unaffected and continue to function. Because a completed proposal is submitted through the ordinary multi-sign path, the multi-sign validation rules are unchanged. The one addition to the common path is a cleanup check: when any account consumes a `TicketSequence`, the ledger removes a matching `TransactionProposal` if one exists (§4.5). Accounts that do not use On-Chain Cosigner are not impacted.
```
ckeshava: As per the earlier comment, this section could use a mention of the `terTICKET_RESERVED` error case.

shawnxie999: https://github.com/XRPLF/XRPL-Standards/pull/613/commits/fe7ca91c2e287a157100372baac60a1dd2490430

XLS-draft-onchain-cosigner/README.md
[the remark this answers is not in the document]
shawnxie999: https://github.com/XRPLF/XRPL-Standards/pull/613/commits/fe7ca91c2e287a157100372baac60a1dd2490430

XLS-draft-onchain-cosigner/README.md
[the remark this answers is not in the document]
shawnxie999: Fixed https://github.com/XRPLF/XRPL-Standards/pull/613/commits/fe7ca91c2e287a157100372baac60a1dd2490430

XLS-draft-onchain-cosigner/README.md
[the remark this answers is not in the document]
shawnxie999: https://github.com/XRPLF/XRPL-Standards/pull/613/commits/fe7ca91c2e287a157100372baac60a1dd2490430

XLS-draft-onchain-cosigner/README.md
ckeshava, line 1158: I believe it is helpful to add an advisory / FAQ, with regards to Proposals containing Batch transactions. Suppose batch-inner transactions containing `Sequence` numbers. An unrelated transaction that consumes this inner-batch-transaction's Sequence could invalidate the entire proposal (assuming the `tfAllOrNothing` flag is used on the Batch transaction)

Alternatively, if the community feels it is common practice to use Tickets along with Batch transactions, we can ignore this  suggestion.

ckeshava, line 128: I recently observed that `sfSigningPubKey` does not carry the `SField::kNotSigning` tag. This implies that `sfSigningPubKey` is not considered a signing-field by the XRPL code.

In such transactions as those which need multiple signatures (Batch, LoanSet, Sponsor-sig), the order of signature-collection changes results. All the signatures collected before the specification of `sfSigningPubKey` will sign-over a different data-payload, compared to the signatures collected after the `sfSigningPubKey` is set.

Here is a test that makes use of the `jtx::sign` method. I have not yet verified this behavior against the proposed TransactionProposalSign transaction:

```
void
testAuxSignatureSurvivesLaterTopLevelSingleSign()
{
    testcase("aux signature survives a later top-level single-signature");
    using namespace test::jtx;

    Env env{*this, testable_amendments()};
    Account const alice{"alice"};
    Account const sponsor{"sponsor"};
    Account const dest{"dest"};
    env.fund(XRP(10'000), alice, sponsor, dest);
    env.close();

    // The canonical form a TransactionProposal stores while collecting
    // (README.md:133): unsigned, SigningPubKey present and empty.
    auto const buildPayload = [&]() {
        json::Value jv = pay(alice, dest, XRP(1));
        jv[jss::Sequence] = env.seq(alice);
        jv[jss::Fee] = "20";
        jv[jss::SigningPubKey] = "";
        jv[sfSponsor.jsonName] = sponsor.human();
        jv[sfSponsorFlags.jsonName] = spfSponsorFee;
        return jv;
    };

    // ---- Order A: SigningFor = Sponsor contributes first. ----
    {
        json::Value jv = buildPayload();

        json::Value sponsorSig{json::ValueType::Object};
        jtx::sign(jv, sponsor, sponsorSig);
        jv[sfSponsorSignature.jsonName] = sponsorSig;

        // alice single-signs for herself: README.md:360 writes this into
        // the payload's TOP-LEVEL SigningPubKey/TxnSignature.
        jtx::sign(jv, alice);

        // §6.5: the payload "is a valid signed transaction that anyone can
        // copy and submit"; §13.3 guarantees it passes validation.
        STTx const stx{jtx::parse(jv)};
        auto const result = stx.checkSign(env.current()->rules());
        BEAST_EXPECTS(result.has_value(), result ? "" : result.error());
    }

    // ---- Order B (control): the account contributes first. ----
    // Identical contributions, opposite arrival order — this one verifies.
    {
        json::Value jv = buildPayload();
        jtx::sign(jv, alice);

        json::Value sponsorSig{json::ValueType::Object};
        jtx::sign(jv, sponsor, sponsorSig);
        jv[sfSponsorSignature.jsonName] = sponsorSig;

        STTx const stx{jtx::parse(jv)};
        BEAST_EXPECT(stx.checkSign(env.current()->rules()).has_value());
    }
}
```

XLS-draft-onchain-cosigner/README.md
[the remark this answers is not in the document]
shawnxie999: fixed

XLS-draft-onchain-cosigner/README.md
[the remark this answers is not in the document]
shawnxie999: updated, although some checks can be done in preflight, but majority of the verifcation is done in preclaim

XLS-draft-onchain-cosigner/README.md
shawnxie999: Let's add this as separate PR, the current spec in this PR is not meant to be finalized, but to serve as a baseline to iterate. (also unblock merging into develop for rippled)

shawnxie999: Good catch! You are right. I believe the main problem is due to supporting single signing, which requires `sfSigningPubKey` to be non-empty. So in the case that when the txn is single signed, the pub key has to be included in the payload. 

However note that for lending/sponsor transactions, `CounterpartySignature` or `SponsorSignature` are not signing fields, so they don't introduce ordering problem, but worth double checking.

Here are a few possible scenarios off the top of my head:
1. Single sign a normal txn - shouldn't matter
2. Single sign a Loanset or sponsored txn. In that case, the signature of `Account`/`Delegate` needs to arrive before `CounterpartySignature` or `SponsorSignature`
3. Single signing outer batch - i think it won't be a problem. Since signatures BatchSigners don't sign the outer payload - they sign the inner txn hashes (but should be verified)

Kassaking7 (approved):
LGTM as the first draft

mvadari:
@xrplf-bot

mvadari:
@ai-review --aggressive

XLS-draft-onchain-cosigner/README.md
ckeshava: To expand on the second point, it can be split into the below four cases. These cases need explicit test cases to verify their behavior, under the `TransactionProposalSign` feature.

image

I believe we can update the spec with further changes, during the development of the `TransactionProposalSign` portion of the feature.

ckeshava (approved):

shawnxie999:
@mvadari I think this is ready to merge

mvadari:
@ai-review --aggressive

XLS-draft-onchain-cosigner/README.md
```diff
@@ -0,0 +1,1201 @@
...
+| `PreviousTxnLgrSeq`   | No       | Yes      | UINT32        | N/A                   | Ledger sequence of the previous transaction that modified this object.                                                                                                                                                                               |
+
+**Field Details:**
+
+#### 4.2.1. `ProposedTransaction`
+
+`ProposedTransaction` is the proposed transaction the proposal collects signatures for. Every field of it is **immutable** for the life of the proposal **except its signature fields** — the top-level `SigningPubKey`/`TxnSignature` (filled only when the target account signs with its own key, §6.1.2); `Signers`; the auxiliary co-signature field(s) a type requires (`CounterpartySignature`, `SponsorSignature`); and, for a `Batch`, `BatchSigners` — into which the ledger inserts each validated signature (see §4.2.2). Because the XRPL signing payloads exclude these signature fields, appending a signature never changes what any signer signed over, so previously-collected signatures stay valid and later signers sign the same canonical payload.
```
mvadari: [bot] **blocking** — The top-level `SigningPubKey` is a signing field, so "the XRPL signing payloads exclude these signature fields, so appending a signature never changes what any signer signed over" does not hold for it. In rippled, `sfTxnSignature`, `sfSigners`, and `sfBatchSigners` carry `SField::kNotSigning`; `sfSigningPubKey` does not (`include/xrpl/protocol/detail/sfields.macro:289-290,429,455`), and `getSigningData` serializes via `addWithoutSigningFields` (`src/libxrpl/protocol/STTx.cpp:171`).

Concretely: a proposed `LoanSet` whose lender co-signs first signs a payload with `SigningPubKey: ""`. If the target account then single-signs per §6.1.2, the top-level `SigningPubKey` is filled, the serialized signing data changes, and the already-collected `CounterpartySignature` no longer verifies — the completed transaction fails `temBAD_SIGNER` at submission. That contradicts this section's "previously-collected signatures stay valid" and §A.10's "the parties can sign in any order".

Either specify that a single-sign contribution's payload is computed with `SigningPubKey` set to the contributed key and that it must therefore be the first/only contribution, or drop the top-level single-sign path and route the target account's own key through `Signers`.

XLS-draft-onchain-cosigner/README.md
```diff
@@ -0,0 +1,1201 @@
...
+2. The proposed transaction fails the **stateless format checks (preflight) for its own transaction type**. These are the same checks it would receive if submitted directly, except for signature-presence and signature-verification checks because the payload is intentionally unsigned (§4.2.1). If any check fails, the proposal returns the `tem` code from that transaction type's preflight. Running these checks at creation is cheap and rejects malformed payloads immediately, instead of letting an invalid proposal gather signatures only to fail later. State-dependent (preclaim) checks are **not** run here; they are evaluated when the completed transaction is submitted.
+3. The proposed transaction is not unsigned. The result depends on the signature field and its contents:
+   - A non-empty `TxnSignature`; a non-empty `SigningPubKey` combined with a `Signers` array; or `Signers` entries carrying real signatures returns `temINVALID`.
```
mvadari: [bot] **blocking** — No condition here rejects a non-empty but well-formed `SigningPubKey` at creation. This bullet's `SigningPubKey` clause is unreachable, because §4.2.1 requires `Signers` to be omitted at creation, and bullet 4 only catches an invalid key type.

So a proposal can be created carrying a valid `SigningPubKey` and no signatures — permitted by every listed condition, forbidden by §4.2.1 and the §4.6 invariant. Multi-sign shares then accumulate into `Signers`, the proposal reports `complete`, and submission dispatches on `signingPubKey.empty()` (`src/libxrpl/protocol/STTx.cpp:258-261`) into `checkSingleSign`, which rejects with "Cannot both single- and multi-sign". A `complete` proposal that cannot be submitted contradicts §1, §6.5, and §13.3.

Add an unconditional "a non-empty `SigningPubKey` returns `temINVALID`" (as already done for `SponsorSignature`) and reconcile bullet 4, which becomes unreachable.

XLS-draft-onchain-cosigner/README.md
```diff
@@ -0,0 +1,1201 @@
...
+### 7.4. Failure Conditions
+
+#### 7.4.1. Data Verification
+
+1. `ProposalID` is missing or malformed (`temMALFORMED`).
+
+#### 7.4.2. Protocol-Level Failures
+
+1. No `TransactionProposal` object exists with the given `ProposalID` (`tecNO_ENTRY`).
+2. The proposal is not terminal and `Account` is neither the `Owner` nor the target account (`tecNO_PERMISSION`).
```
mvadari: [bot] **should-fix** — §7.2 authorizes "the target account (the proposed transaction's `Account` or `Delegate`)" to cancel, but this condition only exempts the `Owner` and "the target account", which §3.1 defines as the `Account` alone. A `Delegate` cancelling a delegated proposal is authorized by §7.2 and refused with `tecNO_PERMISSION` here. §4.5's deletion conditions omit the `Delegate` too.

```suggestion
2. The proposal is not terminal and `Account` is neither the `Owner` nor the target account — or its `Delegate`, when the proposed transaction has one — (`tecNO_PERMISSION`).
```

XLS-draft-onchain-cosigner/README.md
```diff
@@ -0,0 +1,1201 @@
...
+| `TransactionType` | ✔️        | string    | UINT16        | `TransactionProposalCancel` | Identifies this as a `TransactionProposalCancel` transaction. |
+| `Account`         | ✔️        | string    | ACCOUNT       | N/A                         | The account requesting cancellation.                          |
+| `ProposalID`      | ✔️        | string    | HASH256       | N/A                         | The ID of the `TransactionProposal` to cancel.                |
+
+### 7.2. Authorization
+
+- **Non-terminal proposal:** the **owner** (the proposal's `Owner`, i.e. the proposer) or the **target account** (the proposed transaction's `Account` or `Delegate`) may cancel.
+- **Terminal proposal:** **Any** account may cancel, to clean up the object and release the owner's reserve.
+
+The target account can cancel at any point in the lifecycle — even after the proposal is complete — without owning the object. Since anyone can create a proposal against any account, and doing so reserves one of that account's tickets (§5.4), the target account needs a way to refuse; cancelling clears the proposal and frees the ticket.
```
mvadari: [bot] **should-fix** — "anyone can create a proposal against any account" contradicts §5.1.1 and §13.6 ("Proposals cannot come from arbitrary accounts… Squatting is therefore an insider risk, not an open one"). §13.6 was corrected earlier in this PR; this copy is stale, and it justifies the target account's cancel authority with a threat that the spec rules out. The cross-reference is also wrong — the ticket reservation is defined in §4.2.1, not §5.4.

```suggestion
The target account can cancel at any point in the lifecycle — even after the proposal is complete — without owning the object. Since any of its signers — or a `Delegate`, when the proposed transaction has one — can create a proposal against it, and doing so reserves one of that account's tickets (§4.2.1), the target account needs a way to refuse; cancelling clears the proposal and frees the ticket.
```

XLS-draft-onchain-cosigner/README.md
```diff
@@ -0,0 +1,1201 @@
...
+6. The contribution is already recorded — `ProposalSignature.Account` is already present in that destination, or a single-signature entry for `SigningFor` already exists (`tecDUPLICATE`). (The same `ProposalSignature.Account` may still sign for a different `SigningFor`, or for a different slot of the same `SigningFor` under a different payload.)
+7. The contribution conflicts with the existing authorization mode for `SigningFor` — a multi-signature share when a single-signature entry is already recorded, or vice versa (`tecNO_PERMISSION`).
+8. Adding the share would exceed the maximum of 32 entries in the destination `Signers` array, or would add a `BatchSigner` past the 24-entry `BatchSigners` limit (`tecOVERSIZE`).
+9. The contribution would leave the proposed transaction's `Fee` below the minimum for the signatures it would then carry (§4.2.1) (`tecINSUFFICIENT_FEE`). Checked on every contribution, so a proposal never collects more signatures than its fee pays for.
```
mvadari: [bot] **should-fix** — `tecINSUFFICIENT_FEE` does not exist. `include/xrpl/protocol/TER.h` on `develop` defines `tecINSUFF_FEE = 136`, `telINSUF_FEE_P`, and `terINSUF_FEE_B` — there is no `tecINSUFFICIENT_FEE`. Use `tecINSUFF_FEE`, or state why the existing codes are insufficient for a new one.

XLS-draft-onchain-cosigner/README.md
```diff
@@ -0,0 +1,1201 @@
...
+
+**Fee Structure:** Standard. The submitter pays the standard fee for this transaction. The proposed transaction's own `Fee` is not charged here; it is charged to that transaction's own fee payer when the completed transaction is submitted (§4.2.1, §13.7).
+
+### 6.3. Failure Conditions
+
+#### 6.3.1. Data Verification
+
+All Data Verification failures return a `tem`-level error.
+
+1. `ProposalID` is missing or malformed (`temMALFORMED`).
```
mvadari: [bot] **should-fix** — This section returns `temMALFORMED` when a required field is absent, but §5.3.1 #1 and #7 state the opposite model for the same case: "required by the transaction format, so if it is missing, deserialization fails before preflight; submission returns a parse error rather than a transaction result, and no fee is charged". `ProposalID`, `SigningFor`, and `ProposalSignature` are all marked required in §6.1 (same for `ProposalID` in §7.4.1 #1), so implementers get two different error surfaces — and two different fee outcomes — for one situation. Apply §5.3.1's wording here and in §7.4.1, keeping `temMALFORMED` for present-but-malformed values and for absent sub-fields of `ProposalSignature`.

XLS-draft-onchain-cosigner/README.md
```diff
@@ -0,0 +1,1201 @@
...
+
+- `Expiration`: the queried ledger's parent close time has reached or passed the proposal's `Expiration`.
+- `LastLedgerSequence`: the proposed transaction can no longer be included in any ledger. The earliest ledger it could still enter is the queried ledger itself when that ledger is open, and the next ledger otherwise — so the proposal is expired when `LastLedgerSequence` is below that bound. (A transaction with `LastLedgerSequence` equal to the current open ledger's sequence is still submittable and reports `pending`/`complete`, not `expired`.)
+
+##### 8.1.3.4. `complete` Is an Authorization Verdict
+
+`complete` asserts that every required authorization is satisfied on the queried ledger — no more. It does not re-validate everything submission will: the ticket the proposal is keyed on may never have been created (`tefNO_TICKET`), the target account may since have been deleted, the fixed `Fee` may be unfundable, or an amendment the transaction needs may have been disabled. Clients should treat `complete` as "assemble and submit now, and expect success under normal conditions", not as a guarantee of `tesSUCCESS`.
```
mvadari: [bot] **should-fix** — "the ticket the proposal is keyed on may never have been created (`tefNO_TICKET`)" is unreachable: §5.3.2 #7 rejects creation unless the `TicketSequence` is a valid `Ticket` of the target account, and §4.2.1 reserves that ticket against consumption for the proposal's lifetime. As written it tells client authors to defend against a state the spec prevents. Drop it, or replace it with a reachable example — e.g. the fixed `Fee` no longer meeting the required minimum after fee escalation.

XLS-draft-onchain-cosigner/README.md
```diff
@@ -0,0 +1,1201 @@
...
+- **Must** specify a `TicketSequence` for its target account and **must not** specify `Sequence`. Requiring a ticket decouples the proposed transaction from the target account's live sequence, so unrelated target-account activity cannot invalidate the proposal while signatures are being collected (see §9.2). While the proposal exists, that ticket is **reserved**: only the proposed transaction may spend it, and any other transaction that tries fails with `terTICKET_RESERVED`. A transaction counts as the proposed transaction when every field except the signature fields matches `ProposedTransaction`, so the verbatim copy from §6.5 qualifies even if the submitter drops surplus signatures. Deleting the proposal (§4.5) lifts the reservation, which is why the code is a retriable `ter`.
```
mvadari: [bot] **should-fix** — `terTICKET_RESERVED` is a new result code. This sentence justifies the `ter` *class* ("Deleting the proposal lifts the reservation, which is why the code is a retriable `ter`") but not the new code; `TER.h` already has `terPRE_TICKET` for a ticket that is not yet usable. Add one sentence on why the existing `ter` codes do not cover this.

mvadari (commented):
[bot] Conformity skill run against ace222c. Validator clean (`validate_xls_template.py` ✓, `xls_parser.py` ✓ 81 docs) — nothing CI owns below. Two blocking findings, both on the top-level `SigningPubKey` handling; the rest are cross-reference and error-code defects. See inline.

shawnxie999:
@ai-review --aggressive

shawnxie999:
@mvadari resolved the comments

mvadari (commented):
[bot] Second pass, re-baselined on 1be806e. All eight findings from my ace222c review are fixed — thanks. Validator still clean. Six remaining: five that the previous pass turned up in untouched sections, plus one introduced by the new first-signature rule. One blocking. See inline.

XLS-draft-onchain-cosigner/README.md
```diff
@@ -0,0 +1,1204 @@
...
+- Re-checks the proposed transaction's `Fee` against the signatures it now carries (§4.2.1), so the declared fee always covers the whole set.
+- No execution occurs. Once the collected signatures satisfy every signing requirement for the proposed transaction — the target account's quorum, plus a satisfied signature for each `Counterparty`/`Sponsor` the transaction requires, or, for a `Batch`, the outer account's quorum plus a satisfied authorization for every participant account — the proposal is **complete**: the `ProposedTransaction` field is a valid signed transaction that anyone can copy and submit (see §6.5).
+
+**On failure against a terminal proposal (`tecEXPIRED`):**
+
+- No signature is recorded. Because a `tec` result is still applied to the ledger, the terminal proposal object is deleted and the `Owner`'s `OwnerCount` is decremented by 5, or 10 for a `Batch` (§4.4), releasing the reserve as a side effect. This mirrors how transactions like `EscrowFinish` and `CheckCash` report a claimed-fee failure while cleaning up an expired object. A signer whose `TransactionProposalSign` arrives after the proposal has expired therefore both fails and cleans up in one step.
```
mvadari: [bot] **blocking** — This side effect is not implementable as specified, and neither cited precedent behaves this way.

§6.3.2 lists the terminal check under Protocol-Level Failures, i.e. preclaim. A `tec` from preclaim never reaches `doApply` (`src/libxrpl/tx/Transactor.cpp:1589-1591`: `auto result = ctx_.preclaimResult; if (isTesSuccess(result)) result = apply();`), so there is no deletion to preserve. `processPersistentChanges` only re-applies deletions `doApply` already made, and its `tecEXPIRED` allow-list is hard-coded to `ltNFTOKEN_OFFER` and `ltCREDENTIAL` (`Transactor.cpp:1440-1456`), under a FIXME describing the mechanism as "awkward and very limiting".

The precedents do the opposite of what this paragraph claims: `CheckCash::preclaim` returns `tecEXPIRED` and leaves the Check in place (`CheckCash.cpp:133-136`; the `erase` in that file is the success path), and `EscrowFinish`'s `CancelAfter` check returns `tecNO_PERMISSION` without deleting the Escrow (`EscrowFinish.cpp:259-260`).

Result: a late signer's `TransactionProposalSign` fails with `tecEXPIRED` and the proposal survives with its reserve still held, contradicting this bullet and §4.5's "Incidental cleanup by a late signer".

Either state that the terminal check runs in `doApply` and that `TransactionProposal` must be added to the `tecEXPIRED` persistent-deletion set, or drop the side effect and leave cleanup to `TransactionProposalCancel`.

XLS-draft-onchain-cosigner/README.md
```diff
@@ -0,0 +1,1204 @@
...
+| `ProposalSignature` | ✔️        | object    | STOBJECT      | N/A                       | The contributed signature: `Account`, `SigningPubKey`, and `TxnSignature` over `SigningFor`'s signing data (§6.1.2).                 |
+
+`ProposalSignature` contains three required fields: `Account`, `SigningPubKey`, and `TxnSignature`. `ProposalSignature.Account` is the signer account ID used for authorization checks and for the stored signature entry. The outer transaction `Account` can be any funded account willing to submit and pay the fee.
+
+#### 6.1.1. `SigningFor` — which account is being authorized, and where the signature lands
+
+`SigningFor` answers the question: **whose approval is this signature providing?** It must name an account that the **proposed transaction** needs a signature from. In other words, `SigningFor` must point to one of the proposed transaction's signature slots:
+
+- the proposed transaction's own `Account`, or its `Delegate` if permission delegation is used;
```
mvadari: [bot] **should-fix** — `Account` and `Delegate` are alternatives here and in §6.3.2 #3, but they must be mutually exclusive. When `sfDelegate` is present, `Transactor::checkSign` validates the transaction's `Signers` against the Delegate alone (`Transactor.cpp:1011-1015`), so shares admitted under `SigningFor` = the target `Account` land in `ProposedTransaction.Signers` and void the whole set at submission (§8.1.3.2). Because signatures only accumulate, the proposal is then permanently unsatisfiable and has to be cancelled.

§6.3.2 #3 already applies exactly this rule to Batch inners ("an inner account that its `Delegate` replaces… is refused here instead of failing at submission"); the top-level case needs the same treatment, matching §8.1.3.1's exclusive `account` row.

XLS-draft-onchain-cosigner/README.md
```diff
@@ -0,0 +1,1204 @@
...
+5. `ProposalSignature.TxnSignature` is not valid over any signing payload `SigningFor` owes for the stored proposed transaction (§6.1.1, §6.1.2) (`tefBAD_SIGNATURE`). This is the stateful half of §6.3.1.3 — the payload is only available once `ProposalID` has been resolved. The contribution is recorded in the slot whose payload it verifies against.
+6. The contribution is already recorded — `ProposalSignature.Account` is already present in that destination, or a single-signature entry for `SigningFor` already exists (`tecDUPLICATE`). (The same `ProposalSignature.Account` may still sign for a different `SigningFor`, or for a different slot of the same `SigningFor` under a different payload.)
+7. The contribution conflicts with the existing authorization mode for `SigningFor` — a multi-signature share when a single-signature entry is already recorded, or vice versa (`tecNO_PERMISSION`).
```
mvadari: [bot] **should-fix** — #6 and #7 prescribe different result codes for one case. A multi-signature share arriving when a single-signature entry already exists for `SigningFor` matches #6's second clause (`tecDUPLICATE`) and #7's first clause (`tecNO_PERMISSION`). Narrow #6 to same-mode duplicates — `ProposalSignature.Account` already present in the destination, or a second single-signature for the same slot — and leave every mode conflict to #7.

XLS-draft-onchain-cosigner/README.md
```diff
@@ -0,0 +1,1204 @@
...
+- **Must** carry a `Fee` big enough for all the signatures the proposal will collect. The payload is immutable, so the fee is fixed before anyone signs, but the minimum fee grows with each signature: `(1 + |Signers|) × base_fee` for a multi-signed transaction, plus `base_fee` per auxiliary signature ([XLS-66 §3.8.1.1](../XLS-0066-lending-protocol/README.md)), and `(n + 2) × base_fee + Σ inner fees` for a `Batch` carrying `n` signatures ([XLS-56 §2.2](../XLS-0056-batch/README.md#22-transaction-fee)). The `Fee` is therefore a **signature budget**: `TransactionProposalSign` rejects any contribution that would push the minimum above it (§6.3.2), which is what keeps a complete proposal submittable verbatim (§6.5). Declare a larger `Fee` to leave room for extra signers. The fee is paid by the transaction's normal fee payer: usually the **target account** (the proposed transaction's `Account`), but the `Delegate` if the transaction is delegated ([XLS-75](../XLS-0075-permission-delegation/README.md)), or the `Sponsor` if its fee is sponsored ([XLS-68](../XLS-0068-sponsored-fees-and-reserves/README.md)). See §13.7.
```
mvadari: [bot] **should-fix** — "plus `base_fee` per auxiliary signature" is wrong for a single-signing `Sponsor`. XLS-68 gives `(1 + |tx.Signers| + |tx.SponsorSignature.Signers|) × base_fee` (XLS-0068-sponsored-fees-and-reserves/README.md:467), and rippled's `Transactor::calculateBaseFee` counts only the nested `SponsorSignature.Signers` (`Transactor.cpp:468-473`) — a sponsor signing with its own key adds nothing to the minimum.

So on a sponsored `Payment` with a two-member multi-sign target, the true minimum is `(1 + 2) × base_fee`, but §6.3.2 #10 computes one base fee more and rejects the sponsor's contribution on a proposal whose `Fee` is exactly submittable. XLS-66's `max(1, n)` rule does charge for a single-signing counterparty, so only the sponsor branch is affected.

Split the auxiliary term: `max(1, |CounterpartySignature.Signers|)` for a `Counterparty` (XLS-66 §3.8.1.1), `|SponsorSignature.Signers|` for a `Sponsor` (XLS-68).

XLS-draft-onchain-cosigner/README.md
```diff
@@ -0,0 +1,1204 @@
...
+
+A proposal exists as a ledger object only until it is cancelled or cleaned up. Its full history — creation, each signature and the ledger it was recorded in, and the final outcome — remains permanently available in transaction metadata for compliance, audit, and reconciliation.
+
+### 3.3. Design principles
+
+- **The ledger is the meeting room, not the executor.** Signatures are collected and validated on-ledger; execution reuses the existing multi-sign submission path. No new execution semantics and no fourth transaction are introduced.
+- **Authority derives from the collected signatures, not from any submitter.** Anyone can submit the completed transaction; the existing multi-sign machinery validates it against the target account's `SignerList`.
+- **Immutable payload.** Once created, the proposed transaction cannot be modified. Signers sign exactly what they see.
+- **No new quorum model.** Quorum, weights, and per-transaction-type signer lists are inherited from [XLS-49 (Multiple Signer Lists)](../XLS-0049-multiple-signer-lists/README.md) and the existing multi-sign machinery.
```
mvadari: [bot] **should-fix** — The spec's normative text depends on XLS-49 throughout — "the applicable `SignerList` for the proposed transaction type" governs §3.1, §4.2.2, §5.1.1, §6.1.2, §6.3.2 #4, §9.3, and §10 — but XLS-49 is not implemented: `Transactor::checkMultiSign` reads the single `keylet::signerList(id)` and asserts `SignerListID == 0`, commented "We have plans to support multiple SignerLists in the future" (`Transactor.cpp:1060-1075`). §11 gates the feature on the `Cosigner` amendment alone, and the preamble carries no `requires:`.

The earlier thread's answer — "XLS49 is not a feature yet" — argues for dropping the dependency, but the text still rests on it in every section above, so an implementer targeting today's ledger has no defined meaning for "applicable". Add `requires:` to the preamble, and say in §11 what "applicable `SignerList`" means while XLS-49 is inactive (the account's single list).

XLS-draft-onchain-cosigner/README.md
```diff
@@ -0,0 +1,1204 @@
...
+6. The contribution is already recorded — `ProposalSignature.Account` is already present in that destination, or a single-signature entry for `SigningFor` already exists (`tecDUPLICATE`). (The same `ProposalSignature.Account` may still sign for a different `SigningFor`, or for a different slot of the same `SigningFor` under a different payload.)
+7. The contribution conflicts with the existing authorization mode for `SigningFor` — a multi-signature share when a single-signature entry is already recorded, or vice versa (`tecNO_PERMISSION`).
+8. The contribution is a top-level single-signature for the proposed transaction's own `Account`/`Delegate` (§6.1.2), and any signature — a `Signers` entry, a `CounterpartySignature`, a `SponsorSignature`, or a `BatchSigners` entry — has already been recorded on the proposal (`tecNO_PERMISSION`). Filling the top-level `SigningPubKey` changes the signing payload every other slot signs over, so this contribution is only accepted as the proposal's first signature; a target account signing later must contribute a `Signers` entry instead.
```
mvadari: [bot] **should-fix** — The new first-signature rule is unrecoverable for a single-key target account, and stricter than the payload requires.

A target account with no `SignerList` cannot contribute a `Signers` entry at all — §6.3.2 #4 requires `ProposalSignature.Account` to be on `SigningFor`'s applicable list, and submission would return `tefNOT_MULTI_SIGNING` (`Transactor::checkMultiSign`). So if any co-signer signs first, the single-key target's only authorization path is closed and the proposal is permanently unsatisfiable, with nothing in §8.1.2 reporting that dead state.

The rule is also unnecessary for `BatchSigners`. The XLS-56 batch payload is `HashPrefix::Batch` + outer `Account` + outer sequence + `Flags` + inner transaction hashes ([XLS-56 §2.1.3.2](../XLS-0056-batch/README.md#2132-signing-payload); rippled `STTx::checkBatchSingleSign`, `STTx.cpp:470-477`) and never covers the top-level `SigningPubKey`, so an unsponsored `Batch`'s outer account can single-sign at any point without disturbing a single participant signature.

Bind the rule to the slots whose payload covers the whole serialized transaction — `Signers`, `CounterpartySignature`, `SponsorSignature` — exempt `BatchSigners`-only proposals, and state outright that a single-key target account must sign first.

mvadari (commented):
[bot] Third pass, re-baselined on fb8c1e1. All six findings from the 1be806e review are fixed — the `doApply` correction, the `BatchSigners` exemption, and the split fee formula all look right. Validator clean. Three remaining, all cross-spec, none blocking. See inline.

XLS-draft-onchain-cosigner/README.md
```diff
@@ -0,0 +1,1204 @@
...
+
+### 5.3. Failure Conditions
+
+#### 5.3.1. Data Verification
+
+Except for missing required fields, Data Verification failures return a `tem`-level error.
+
+1. If present, `ProposedTransaction` is not a well-formed transaction of a known type (`temMALFORMED`). `ProposedTransaction` is required by the transaction format, so if it is missing, deserialization fails before preflight; submission returns a parse error rather than a transaction result, and no fee is charged.
+2. The proposed transaction fails the **stateless format checks (preflight) for its own transaction type**. These are the same checks it would receive if submitted directly, except for signature-presence and signature-verification checks because the payload is intentionally unsigned (§4.2.1). If any check fails, the proposal returns the `tem` code from that transaction type's preflight. Running these checks at creation is cheap and rejects malformed payloads immediately, instead of letting an invalid proposal gather signatures only to fail later. State-dependent (preclaim) checks are **not** run here; they are evaluated when the completed transaction is submitted.
```
mvadari: [bot] **should-fix** — "except for signature-presence and signature-verification checks" cuts both ways, and each reading breaks something.

Read literally, it drops checks that have to stay. XLS-56 §2.3 rules 2.5–2.7 — an inner transaction carrying `TxnSignature` (`temBAD_SIGNATURE`), a non-empty inner `SigningPubKey` (`temBAD_REGKEY`), an inner `Signers` array (`temBAD_SIGNER`) — are literally signature-presence checks, yet they are exactly what enforces §4.2.1's "each unsigned, with the `tfInnerBatchTxn` flag". Dropping them admits a proposed `Batch` with pre-signed inners, which then collects signatures and 10 reserve increments and fails `tem` at submission: an unsubmittable proposal by construction.

Read narrowly, no `Batch` or `LoanSet` proposal can ever be created. XLS-56 §2.3 rule 3.6 (`BatchSigners` missing a required signer, "including the case where the `BatchSigners` field is required but is not present at all") and XLS-66's "`CounterpartySignature` is not present and the transaction is not part of a `Batch` inner transaction" (XLS-0066-lending-protocol/README.md:1049) both return `temBAD_SIGNER` on an intentionally unsigned payload — and those are the two motivating use cases in §10.

Draw the distinction that matters instead: skip checks that require a signature to be **present**, keep checks that require signature fields to be **absent**. Naming the XLS-56 and XLS-66 conditions on each side removes the ambiguity.

XLS-draft-onchain-cosigner/README.md
```diff
@@ -0,0 +1,1204 @@
...
+**Directory Registration:** The object is registered in the `Owner`'s owner directory.
+
+### 4.4. Reserves
+
+**Reserve Requirement:** Custom (flat). A `TransactionProposal` holds a full transaction plus its collected signatures, so it reserves more than a typical ledger entry.
+
+- **Ordinary proposed transaction:** 5 owner-reserve increments (currently **1 XRP**).
+- **`Batch` proposed transaction:** 10 owner-reserve increments (currently **2 XRP**), reflecting its larger footprint (up to 8 inner transactions and signatures for multiple participant accounts).
+
+Each increment is the standard owner-reserve amount (currently 0.2 XRP, subject to Fee Voting). There is no separate reserve mechanism: creating a proposal adds 5 (or 10) to the owner's `OwnerCount` instead of 1, and deleting it subtracts the same amount (§5.4, §6.4, §7.5).
```
mvadari, line 162: [bot] **should-fix** — The reserve accounting is stated unconditionally against the `Owner`'s `OwnerCount`, with no account of XLS-68 reserve sponsorship.

`Sponsor`/`SponsorFlags` are transaction common fields; `spfSponsorReserve` is valid on every transaction except `Batch` and pseudo-transactions (XLS-68 §8.3.4); and XLS-68's entry-type allow-list ends with "Any other ledger entry type that contributes to an account's owner reserve" (§4.3.1). A reserve-sponsored `TransactionProposalCreate` is therefore well-formed today, but §4.2's field table has no `Sponsor`, and §5.4/§6.4/§7.5 move only the `Owner`'s count by 5/10.

One implementer ignores the flag and the sponsor pays nothing; another shifts the reserve at creation and then decrements the `Owner`'s count on deletion as specified here, corrupting the counters and tripping rippled's `SponsorshipOwnerCountsMatch` invariant.

Either say the object may carry `Sponsor` and that the 5/10 count applies to `SponsoredOwnerCount`/`SponsoringOwnerCount` on every create and delete path, or add `TransactionProposalCreate` to XLS-68 §8.3.4's `spfSponsorReserve` exclusion list.

XLS-draft-onchain-cosigner/README.md
```diff
@@ -0,0 +1,1204 @@
...
+| `proposal`        | object | The raw `TransactionProposal` ledger object.                                                                                                                                                                                             |
+| `proposal_status` | string | Where the proposal is in its lifecycle: `"pending"`, `"complete"`, or `"expired"` (see below).                                                                                                                                           |
+| `signing_status`  | array  | One entry per required authorization (§8.1.3.1), in a stable order: the `account` row first, then auxiliary co-signers, then batch participants.                                                                                         |
+| `tx_blob`         | string | Present only when `proposal_status` is `"complete"`: the stored `ProposedTransaction` serialized in submit-ready binary form, so a client does not have to reassemble and re-serialize it. Providing it implies nothing beyond §8.1.3.4. |
```
mvadari: [bot] **should-fix** — `tx_blob` is gated on `proposal_status == "complete"`, but §8.1.2 evaluates terminal-first, so a proposal that reached quorum and then expired reports `expired` with no `tx_blob` — precisely the state §13.4 calls the central security consideration, where "those signatures remain valid regardless of whether the proposal object still exists".

A wallet switching on `proposal_status`, as this section instructs, then shows "expired" to an operator for a transaction that anyone can still copy and submit, and has to re-derive completeness by joining `signing_status` against live `SignerList`s — the work the computed fields exist to avoid.

Emit `tx_blob` whenever every `signing_status` row is satisfied, independent of terminality, or add a separate boolean (e.g. `authorized`) that `expired` does not mask.

XLS-draft-onchain-cosigner/README.md
mvadari: This comment is more confusing than the text

XLS-draft-onchain-cosigner/README.md
```diff
@@ -0,0 +1,1205 @@
...
+
+### 3.3. Design principles
+
+- **The ledger is the meeting room, not the executor.** Signatures are collected and validated on-ledger; execution reuses the existing multi-sign submission path. No new execution semantics and no fourth transaction are introduced.
+- **Authority derives from the collected signatures, not from any submitter.** Anyone can submit the completed transaction; the existing multi-sign machinery validates it against the target account's `SignerList`.
+- **Immutable payload.** Once created, the proposed transaction cannot be modified. Signers sign exactly what they see.
+- **No new quorum model.** Quorum and weights are inherited unchanged from the existing multi-sign machinery. "Applicable `SignerList`" means whatever list standard multi-sign resolution uses today — the account's single `SignerList` — and would automatically mean the per-transaction-type list if [XLS-49 (Multiple Signer Lists)](../XLS-0049-multiple-signer-lists/README.md) is later activated; this proposal does not depend on XLS-49 (§10).
```
mvadari: IMO mentioning/discussing XLS-49 in this spec is more confusing than leaving it out. We can always add that support later.

XLS-draft-onchain-cosigner/README.md
```diff
@@ -0,0 +1,1205 @@
+<pre>
+  title: On-Chain Cosigner
+  description: Native on-ledger proposal and multi-signature collection for XRPL transactions.
+  author: Shawn Xie <shawnxie@ripple.com>, Zhiyuan Wang (@Kassaking7), Chenna Keshava B S (@ckeshava), Mayukha Vadari (@mvadari)
```
mvadari: nit: Please be consistent about whether you use emails or Github handles here

XLS-draft-onchain-cosigner/README.md
```diff
@@ -0,0 +1,1205 @@
...
+
+This feature will require an amendment, tentatively titled `Cosigner`.
+
+## 2. Motivation
+
+XRPL multi-sign today has three structural problems, all stemming from the absence of an on-ledger "meeting room" for signers:
+
+1. **Manual assembly and latency.** There is no way to "send" a transaction to another signer through the ledger. The transaction blob must be passed around through external channels (email, Slack, custody tooling). Gathering signatures is slow and error-prone, which makes multi-sign unsuitable for time-sensitive operations.
+
+2. **No auto-fill.** For a single-signer transaction, fields like `Sequence` and `LastLedgerSequence` can be auto-filled at submission time. For multi-sign, every field must be fixed before the first signer signs. If the transaction fails to reach a ledger in time (for example because `LastLedgerSequence` was set too low while waiting for signers), the entire round must be restarted from scratch with all signers.
```
mvadari: I don't think this issue is fixed by this spec?

XLS-draft-onchain-cosigner/README.md
```diff
@@ -0,0 +1,1205 @@
...
+
+**Key Space:** `0x[TBD]`
+
+**ID Calculation Algorithm:**
+
+```
+ProposalID = hash( <TransactionProposal space key>, Account, TicketSequence )
+```
+
+where `Account` and `TicketSequence` are taken from the **proposed transaction**: `Account` is the target account, and `TicketSequence` is the ticket it spends. **Nothing else contributes to the ID** — not the rest of the payload, and not any signature field — so the ID is fixed at creation and never changes as signatures accumulate. This value is the `ProposalID` referenced by `TransactionProposalSign` and `TransactionProposalCancel`.
```
mvadari, line 103: Are txs with `Sequence` instead of `TicketSequence` not allowed in this feature? If so, that's not mentioned before this section

XLS-draft-onchain-cosigner/README.md
```diff
@@ -0,0 +1,1205 @@
...
+| `OwnerNode`           | Yes      | Yes      | UINT64        | N/A                   | Hint for which page this object appears on in the owner directory.                                                                                                                                                                                   |
+| `PreviousTxnID`       | No       | Yes      | HASH256       | N/A                   | Hash of the previous transaction that modified this object.                                                                                                                                                                                          |
+| `PreviousTxnLgrSeq`   | No       | Yes      | UINT32        | N/A                   | Ledger sequence of the previous transaction that modified this object.                                                                                                                                                                               |
+
+**Field Details:**
```
mvadari: Unnecessary

XLS-draft-onchain-cosigner/README.md
```diff
@@ -0,0 +1,1200 @@
+<pre>
+  title: On-Chain Cosigner
+  description: Native on-ledger proposal and multi-signature collection for XRPL transactions.
+  author: Shawn Xie (@shawnxie999), Zhiyuan Wang (@Kassaking7), Chenna Keshava B S (@ckeshava), Mayukha Vadari (@mvadari)
+  category: Amendment
+  status: Draft
+  proposal-from: https://github.com/XRPLF/XRPL-Standards/discussions/589
+  created: 2026-07-14
```
mvadari, line 8: Should have an `updated` date too

XLS-draft-onchain-cosigner/README.md
```diff
@@ -0,0 +1,1200 @@
...
+| `ProposedTransaction` | ✔️        | object    | STOBJECT      | N/A                         | The unsigned proposed transaction (see §4.2.1).                        |
+| `Expiration`          | ✔️        | number    | UINT32        | N/A                         | Ledger close-time after which the proposal stops accepting signatures. |
+
+Standard common fields (`Fee`, `Sequence`, `Flags`, `Memos`, `SourceTag`, signing fields) apply. `Memos` and `SourceTag` MAY be used to attach a reason code or reconciliation identifier to the proposal.
+
+#### 5.1.1. Creation authorization
+
+Only the target account, or one of its signers, may create a proposal — and when `ProposedTransaction.Delegate` is present, the `Delegate` and its signers stand in for the target account. The `Account` submitting `TransactionProposalCreate` must therefore be either:
+
+- the target account itself, or a member of the target account's applicable `SignerList` for the proposed transaction type, when `ProposedTransaction.Delegate` is absent; or
```
mvadari: [bot] **should-fix** — "for the proposed transaction type" no longer has a definition. This commit deleted the XLS-49 references that gave it meaning, and §3.3 now says the applicable list is simply "the account's `SignerList`". An implementer reading this bullet looks for a list keyed to the payload's transaction type, which no ledger provides, so the creation-authorization check has no defined input. Drop the phrase here, on line 253 for the `Delegate`, and in A.1.

XLS-draft-onchain-cosigner/README.md
```diff
@@ -0,0 +1,1200 @@
...
+- **Incidental cleanup by a late signer:** a `TransactionProposalSign` submitted against a terminal proposal **fails** with `tecEXPIRED` — its intended action (recording a signature) cannot happen — but, as a side effect of that claimed-fee result, it deletes the terminal proposal and releases the reserve (see §6.4).
+- **Automatic cleanup when the proposed transaction executes:** the reserved `TicketSequence` can only be consumed by the proposal's own proposed transaction (§4.2.1), so this is the only way a ticket consumption deletes a proposal — running the completed transaction spends the ticket, and the ledger looks up `hash(<space key>, Account, <consumed ticket>)` (§4.1) and deletes the matching proposal, refunding the `Owner`'s reserve.
+
+This removes only the leftover object. Signatures already copied off-ledger stay valid and submittable until the `TicketSequence` is consumed (§13.4).
+
+**Account Deletion Blocker:** Yes. A `TransactionProposal` object must be deleted before its owner account can be deleted.
```
mvadari, line 187: [bot] **should-fix** — This covers the proposal's `Owner`, but not the target account, which can self-delete and destroy the reserved ticket: `ltTICKET` is in rippled's `nonObligationDeleter`, so `AccountDelete` deletes tickets rather than being blocked by them (`src/libxrpl/tx/transactors/account/AccountDelete.cpp:191-200`).

That breaks two claims. §4.2.1 says "only the proposed transaction may spend it… Deleting the proposal lifts the reservation" — an `AccountDelete` ends the reservation without either. And §4.5's automatic cleanup keys on ticket *consumption*, so it never fires: the proposal survives, uncompletable, holding 5–10 reserve increments until `Expiration`, which has no upper bound (§5.3.2 only requires it to be in the future).

§8.1.2's `account_not_found` reason already treats a deleted target as reachable, so the two sections disagree. Either state that a target-account `AccountDelete` destroys a reserved ticket and makes the proposal terminal — and therefore permissionlessly cleanable — or add a live `TransactionProposal` to the target's account-deletion blockers.

XLS-draft-onchain-cosigner/README.md
```diff
@@ -0,0 +1,1200 @@
...
+
+## 12. Open Questions
+
+- **Reducing the initial construction burden:** Can the initial proposed-transaction construction be simplified further, beyond the ticket-based approach in §9.2?
+- **Revocation:** Should there be a first-class way to revoke a completed proposal's signatures on-ledger (beyond consuming the `TicketSequence`), given that cancellation alone does not prevent submission of already-collected signatures (§13.4)?
+- **Recurring / standing orders:** Use Case 7 (recurring allowances and treasury stipends) suggests a proposal could activate a long-lived standing order rather than a one-shot transaction, potentially composing with a Subscriptions primitive. This is out of scope for this spec but noted as a future extension.
```
mvadari: [bot] **should-fix** — "Use Case 7" points at material the document does not contain: it runs §1–§13 plus Appendix A, with no use-case section. A reader cannot evaluate this open question. Describe the case inline in one clause, or link the Discussion it came from.

mvadari (commented):
[bot] Fourth pass, re-baselined on e25d689. The preflight carve-out and `tx_blob` findings are both fixed. Validator clean. Three new: two follow from the XLS-49 removal and the ticket-reservation model. The XLS-68 reserve-sponsorship comment on fb8c1e1 is still open — not repeated here.

XLS-draft-onchain-cosigner/README.md
shawnxie999: No the target account needs to delete the proposal to free up the ticket

XLS-draft-onchain-cosigner/README.md
```diff
@@ -0,0 +1,1200 @@
...
+
+Every `Signers` array (top-level or nested in a `BatchSigner`) is kept sorted by `Account` and holds at most 8 entries (the maximum `SignerList` size). `BatchSigners` is also sorted by `Account` and holds at most 24 entries. **Weights are not stored**: a signer's weight and the relevant quorum are always read from the applicable account's `SignerList`, both when a signature is added and when the transaction is finally submitted (see §9.3). Clients compute "remaining weight to quorum" by joining the collected signatures against the relevant `SignerList`(s). §6.1 describes how `TransactionProposalSign` routes a signature from its `SigningFor` account and `ProposalSignature.Account`.
```
mvadari: [bot] **blocking** — 8 is the pre-`ExpandedSignerList` limit, so both the number and "(the maximum `SignerList` size)" are wrong today. rippled sets `kMaxMultiSigners = 32` (`include/xrpl/protocol/STTx.h:48-49`) and `SignerListSet` enforces `1..32` with no amendment gate (`src/libxrpl/tx/transactors/account/SignerListSet.cpp:249-253`; the comment at :167 reads "entryCount should always be in the range from 1 to 32"). `ExpandedSignerList` is retired — permanently enabled, conditional code removed (`include/xrpl/protocol/detail/features.macro:134`).

An account with 9–32 signers is legal today and can never complete a proposal: §6.3.2 #9 (line 393) rejects the 9th share with `tecOVERSIZE`, and §4.6's invariant (line 192) rejects a 32-entry array that submission accepts.

Restore 32 on lines 147, 192, and 393. If the cap is deliberate rather than a correction, it needs its own justification plus a statement that accounts with more than 8 signers cannot use this feature.

mvadari (commented):
[bot] Fifth pass on 810b0df. The two findings from e25d689 are fixed. One new, blocking: this commit's 32 → 8 change to the max `Signers` size is a pre-`ExpandedSignerList` number. Validator clean.

XLS-0103-onchain-cosigner/README.md
```diff
@@ -0,0 +1,1200 @@
...
+- **Incidental cleanup by a late signer:** a `TransactionProposalSign` submitted against a terminal proposal **fails** with `tecEXPIRED` — its intended action (recording a signature) cannot happen — but, as a side effect of that claimed-fee result, it deletes the terminal proposal and releases the reserve (see §6.4).
+- **Automatic cleanup when the proposed transaction executes:** the reserved `TicketSequence` can only be consumed by the proposal's own proposed transaction (§4.2.1), so this is the only way a ticket consumption deletes a proposal — running the completed transaction spends the ticket, and the ledger looks up `hash(<space key>, Account, <consumed ticket>)` (§4.1) and deletes the matching proposal, refunding the `Owner`'s reserve.
+
+This removes only the leftover object. Signatures already copied off-ledger stay valid and submittable until the `TicketSequence` is consumed (§13.4).
+
+**Account Deletion Blocker:** Yes. A `TransactionProposal` object must be deleted before its owner account can be deleted.
```
mvadari: [bot] The cancel-then-spend path is clear; this is about `AccountDelete`, which deletes tickets rather than spending them — `ltTICKET` is in rippled's `nonObligationDeleter` (`src/libxrpl/tx/transactors/account/AccountDelete.cpp:191-200`). Deleting the target account removes the reserved `Ticket` with no transaction for `terTICKET_RESERVED` to reject, so the proposal survives pointing at a `TicketSequence` that no longer exists and §4.5's cleanup-on-consumption never fires. One sentence saying the proposal becomes terminal in that case would close it.

mvadari (approved):

PR #623, opened 2026-09-08
docs(XLS-66): align fee redirection and DefaultCovered with the normative text

sdh2222:

Two documentation alignments in XLS-66. No protocol change.

1. Section 2.4 currently says the protocol "will not charge the fees" when First-Loss Capital is short. 3.11.5 and 4.6 already specify that the borrower still pays; only the fee destination changes (broker owner vs. cover pool). 2.4 now points at that rule.
2. Appendix A-2 formula (35) defined `DefaultCovered` with two terms. 3.10.5 uses three, including `LoanBroker.CoverAvailable`. The third term is added in the formula, the process list, the 3.1.11 example, and the glossary. The example numbers are unchanged (`CoverAvailable` is not the binding term there).

These are the two doc mismatches only. A separate design question about `LoanBrokerCoverWithdraw` after default is in https://github.com/XRPLF/XRPL-Standards/discussions/622 and is not part of this PR.

### Context of Change

Updating an existing XLS so the overview and the appendix match the normative transaction text. Implementers who start from 2.4 or formula (35) can currently get a different borrower payment or a cover deduction larger than `CoverAvailable`.

### Type of Change

- [x] Documentation (README updates, typo fixes)

## Test plan

- [ ] Confirm 2.4 now matches 3.11.5 item 1 and 4.6 (fees redirected, not waived)
- [ ] Confirm formula (35), the A-2 process step, and the 3.1.11 example all include `CoverAvailable`
- [ ] Confirm the 3.1.11 numeric result is still 10.9 Tokens

Made with [Cursor](https://cursor.com)

### DIFF
--- a/XLS-0066-lending-protocol/README.md
+++ b/XLS-0066-lending-protocol/README.md
@@ -85,7 +85,7 @@ See [Appendix A-3](#a-3-loanpay-implementation-reference) for detailed formulas 
 
 ### 2.4 Fees
 
-The lending protocol charges a number of fees that the Loan Broker can configure. The protocol will not charge the fees if the Loan Broker has not deposited enough First-Loss Capital.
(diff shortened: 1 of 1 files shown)


sdh2222:
Closing this for now — it was opened before we had agreed to publish.

PR #632, opened 2026-09-11
spec(XLS-86): sync the firewall specification with the reference implementation

dangell7:
Brings XLS-86 into line with the reference implementation, which has been ported
onto current `rippled` develop. The implementation change is
Transia-RnD/rippled#38.

Nothing about the firewall's design changes. These are corrections where the
specification and the code disagreed, plus the details the port settled.

## Counterparty signature

`CounterpartySignature` was typed `ARRAY` in all three transaction tables. It is
one `OBJECT`. The object carries `SigningPubKey` and `TxnSignature` for a single
signer, or `Signers` when the counterparty signs with a signer list, so one field
covers both and no array is needed.

This is the field the ledger already uses wherever a second party authorizes
another account's transaction, verified on the common signing path under its own
signature role. Roles sign distinct hash prefixes, so a countersignature cannot
be replayed into a different role or onto a different transaction. New §5.0
records the reasoning.

Consequent result-code corrections: a missing signature is `temBAD_SIGNER` on
`FirewallSet` and `temMALFORMED` on `FirewallDelete` and `WithdrawPreauth`, where
the field is required; a signature not made by the recorded `Counterparty` is
`tefBAD_AUTH`. The rule forbidding the outer account from appearing among the
signers went with the array.

## Corrections

- `WithdrawPreauth` ledger entry type was `0x0856`; it is `0x0086`.
- Transaction type numbers are now stated: `ttWITHDRAW_PREAUTH` 103,
  `ttFIREWALL_SET` 104, `ttFIREWALL_DELETE` 105.
- `PaymentChannelFund` was classified `block`; it is `check`.

## Additions

- New §4.4 states the result code `tefFIREWALL_BLOCK` and its consequence: a
  `tef` is not applied and claims no fee, so a blocked transaction costs the
  protected account nothing.
- §4.4 also states the fee rule — a transaction carrying `CounterpartySignature`
  owes one extra base fee, or one per signer-list entry.
- §4.5, formerly §4.4, replaces the partial transaction table with the full
  classification: `Allow` is the default and covers 41 of 85 types, 11 are
  `Check` and 33 are `Block`, each named. It also records that `FirewallAction`
  is a `TxSettings` member, and warns that because `Allow` is the default, a
  transaction type added later is unguarded until classified.

## Verification

`xrpl.app.Firewall`: 9 cases, 466 tests, 0 failures, covering each rule stated
here.

### DIFF
--- a/XLS-0086-firewall/README.md
+++ b/XLS-0086-firewall/README.md
@@ -171,7 +171,7 @@ SHA512Half(KeySpace || OwnerAccountID || AuthorizedAccountID || DestinationTag)
 
 | Field Name        | Constant | Required | Internal Type | Default Value | Description                                 |
 | ----------------- | -------- | -------- | ------------- | ------------- | ------------------------------------------- |
-| LedgerEntryType   | Yes      | Yes      | UINT16        | 0x0856        | Identifies as WithdrawPreauth               |
+| LedgerEntryType   | Yes      | Yes      | UINT16        | 0x0086        | Identifies as WithdrawPreauth               |
 | Flags             | No       | Yes      | UINT32        | 0             | Reserved for future use                     |
 | Account           | Yes      | Yes      | ACCOUNT       | N/A           | Account that owns this preauth              |
 | Authorize         | Yes      | Yes      | ACCOUNT       | N/A           | Account authorized to receive assets        |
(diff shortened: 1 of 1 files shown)


mvadari:
@dangell7 fyi we're moving to a more AI-heavy process to lessen the load on the human reviewers. We'll take a (human) look once both the AI bots have approved the PR. Let us know if you have any feedback on this process, it's still new and we're still working out kinks in the prompts.

